PR #18, opened 2024-09-27
add section for extra info in config

leifdenby:
Add section called `extra` in config which is ignored by `mllam-data-prep` but can be used to add any extra info the user might want to keep track of. The intention is that this will be used in neural-lam to keep of projection information for now at least


leifdenby:
Just FIY I have decided that this hacky approach (where an `extra` section is added that is ignore by the application) isn't something I feel comfortable adding to `mllam-data-prep`. Instead I have asked @SimonKamuk and @observingClouds to look into implementing #33 so adding projection information can be done properly

leifdenby:
I am reopening this because it is turning out to be currently impossible to add proper WKT-string based projection info to the config and have that parsed and used in `cartopy` to create a `cartopy.crs.Projection` object that can be used for plotting.

I do need to fix that the example I am showing with DANRA had the wrong default globe (as in correct default radius for the globe) relative to what is assumed in DANRA.

I.e. the example in the config of this PR results in the projection being
```
cartopy.crs.LambertConformal(
    central_latitude=56.7, central_longitude=25, standard_parallels=(56.7,56.7), 
)
```
Which has the incorrect (default) globe radius.

whereas it should be (thank you @SimonKamuk!):
```python
cartopy.crs.LambertConformal(
    central_latitude=56.7, central_longitude=25, standard_parallels=(56.7,56.7), 
    globe=cartopy.crs.Globe(semimajor_axis=6367470, semiminor_axis=6367470)
)
```

leifdenby:
Ok, this is ready for review now @observingClouds 🎁 

I added some rudimentary functionality to enable support of multiple schema versions and also wrote up my current plans for schema versioning in the bottom of the README. For now I suggest we inline in the code with comments simply keep track of what schema versions the current version of `mllam-data-prep` supports (in this case it is both `v0.2.0` and `v0.5.0` since the new `extra` field is optional). And I also added tests to ensure that support is there (by keeping copy of of old configs with revisions with should still support).

Hope this makes sense!

PS. I noticed when https://github.com/mllam/mllam-data-prep/pull/35 and https://github.com/mllam/mllam-data-prep/pull/36 were merged in that the changelog entry went in the wrong place for the former and there's no changelog entry for the latter. But I will clean that up when I prepare the `v0.5.0` release. Just in case you felt like the changelog looks quite empty given the git history since the last tag

README.md
```diff
@@ -201,13 +201,25 @@ inputs:
         name_format: "{var_name}"
     target_output_variable: static
 
+extra:
+  projection:
+    class_name: LambertConformal
+    kwargs:
+      central_longitude: 25.0
```
observingClouds, line 208: It might be that we are actually still missing some info here, because from the GRIB files I can see that there is a `false_easting` and `false_northing` different to 0 (the default).

```suggestion
      central_longitude: 25.0
      false_easting: 1999248.19322605
      false_northing: 609541.209132949
```
I have opened an issue with frie data as they also ignore this info in their definition for DINI. @SimonKamuk would be great to have your expertise here.

tests/old_config_schema_examples/v0.2.0/example.danra.yaml
observingClouds, line 1: Should we also keep the old dataclass functions so that one can potentially better convert to/from old schema. The example file helps less with this.

example.danra.yaml
observingClouds, line 1: Should the schema_version here also be adjusted to v0.5.0?

observingClouds (commented):
Looks good to me, the only issue that I see before this can be merged is the correct projection information....once again...

README.md
leifdenby: Ah! I didn't realise we are missing this too.... oh boy. Yes, @SimonKamuk, would be great to know what you think :)

tests/old_config_schema_examples/v0.2.0/example.danra.yaml
leifdenby: I don't fully understand what you mean by "dataclass functions" :) The serialisation to/from yaml is taken care of by `dataclass_wizard`, so there are no functions there, right? And because the new field that I am adding is optional (as in it defaults to `None`) the config files with the old schema can be loaded with the new dataclasses without change. Not sure if this is what you are referring to wrt to conversion though.

example.danra.yaml
leifdenby: Yes! Sorry, I thought I had done that!

leifdenby: done in https://github.com/mllam/mllam-data-prep/pull/18/commits/018ad59c3f76abb515ae042803fe80e7c405912e

README.md
SimonKamuk: image
if I uncomment false_easting and false_northing, the third and fourth assertions fail. So to me it seems either I am doing something wrong or the grib messages are wrong - neither would surprise me 😆

tests/old_config_schema_examples/v0.2.0/example.danra.yaml
observingClouds: Well, I refer to https://github.com/leifdenby/mllam-data-prep/blob/feat/extra-section-in-config/mllam_data_prep/config.py. That file changes with each schema change and is useful to preserve to read/transform from other schema versions.

README.md
observingClouds: wkt_obj_nofalsing_customGlobe
wkt_obj_wfalsing_customGlobe
I went the visual way now and the one that fits, is the one with the northing and easting set to 0.

```
import pyproj
import cartopy
import matplotlib.pyplot as plt
PyProj =  pyproj.crs.CRS.from_wkt(wkt)
CartopyProj = cartopy.crs.Projection(PyProj)
fig, ax = plt.subplots(subplot_kw=dict(projection=ccrs.PlateCarree()))
ds.t0m.isel(time=10).plot(ax=ax, x="x", y="y", transform=CartopyProj)
ax.coastlines()
ax.gridlines(draw_labels=["top", "left"])
fig.show()
```

observingClouds: I am more and more convinced that the grid definition in the GRIB files of DANRA is incorrect, as the smallest x value is -1999248.193226 and the smallest y value is -609541.209133, exactly the numbers of the false_northing and false_easting. Because these numbers are kind of old school and [have been used to avoid negative numbers](https://help.arcgis.com/en/geodatabase/10.0/sdk/arcsde/concepts/geometry/coordref/coordsys/projected/mapprojections.htm), the `false_easting` and `false_northing` should have been subtracted, which obviously did not happen here.

observingClouds: @leifdenby @SimonKamuk thanks for the additional discussion. I feel better now about this definition reflecting the reality to the best of our knowledge 🤣

tests/old_config_schema_examples/v0.2.0/example.danra.yaml
leifdenby: Ah I see, yes so we could keep multiple copies of that file. I was instead thinking we would only keep the current version, but we could resurrect old version later in the development of this package and then write code that does migration between revisions. But I think I will leave this as-is for now :) I hope that is ok, otherwise this is going to start to get very complicated.

README.md
leifdenby: Also, I wonder who we need to inform about these incorrect GRIB definitions? I supposed they will be wrong when we share the GRIB files too... but that is something to work out somewhere else 😄

observingClouds (approved):

leifdenby:
Thanks for all the help @observingClouds and @SimonKamuk! I am merging this in 😄

Issue #22, opened 2024-10-03
Actually make `dask.distributed` optional


leifdenby:
@elbdmi noted that having `dask.distributed` as optional dependency while not handling the situation where it cannot be imported results in an `ImportError` of course. I need to handle this 😄

leifdenby:
closing as this was completed with #27

PR #27, opened 2024-10-03
Actually make dask.distributed optional #22

elbdmi:
- Added a check to make `psutil` and `dask.distributed` optional to improve flexibility.
- Introduced a flag (`dask_distributed_available`) to skip dask-related multiprocessing setup if dependencies are missing.
- Provided a warning message if `dask.distributed` or `psutil` are not available.
- Updated the code to run only dask-specific logic when the necessary modules are present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


elbdmi:
@leifdenby now it is ready

leifdenby (approved):
Perfect! Thank you for doing this :)

Issue #24, opened 2024-10-03
Change config example to call validation split `val` instead of `validation`


leifdenby:
This isn't technically a bug because `mllam-data-prep` doesn't try to enforce any kind of specification wrt what the outputs are called, but the new `datastores` functionality in neural-lam https://github.com/mllam/neural-lam/pull/66 assumes that the validation split is called `val` rather than `validation`, so it would be nicer to give the examples in `mllam-data-prep` to be consistent with that

PR #28, opened 2024-10-21
Call validation split 'val' instead of 'validation' in example config

ealerskans:
The new datastores functionality in neural-lam https://github.com/mllam/neural-lam/pull/66 assumes that the validation split is called `val` rather than `validation`.

With this addition, the example config in `mllam-data-prep` uses `val` instead of `validation` to be consistent with that, as proposed in #24.


observingClouds (approved):
LGTM

leifdenby:
Thanks for doing this @ealerskans and @observingClouds !

Issue #29, opened 2024-10-29
Add ability to derive fields from input datasets


ealerskans:
One outstanding task is to implement the derivation of forcings from `neural-lam` here in `mllam-data-prep`. 

I have had a look at the current forcing calculations from `neural-lam`:
https://github.com/mllam/neural-lam/blob/feature_calculate_forcings/create_forcings.py

All thanks to @observingClouds I have attempted a first draft of the TOA radiation calculations in `mllam-data-prep` here:
- branch: https://github.com/ealerskans/mllam-data-prep/tree/feature/derive_forcings
- commit: https://github.com/ealerskans/mllam-data-prep/commit/981d676ef235313630cadd6985e26de668b4caf1

I have so far only tested on the DANRA zarr files and from this I have some observations:

1. When calculating the TOA radiation we're using: `lat`, `lon`, and `time`, all which are coordinates. Since coordinates are read in eagerly and not lazily in `xarray` we can not just calculate the radiation lazily. However, since the DANRA data is 30 years of data, if we try to just do the calculations using `xarray` we run into memory issues since it would try to allocate ~300 GB. Therefore, Hauke's solution was to create a new dataset using `dask` and chunking and in that way we could do the calculations. The implemented solution is just a first draft and there is some hard-coding included as well. A few points to discuss/iterate over would be:

  - The chunking is currently hard-coded
  ```
  ds_dict = {}
  ds_dict["lat"] = (list(ds.lat.dims), da.from_array(ds.lat.values, chunks=(-1, -1)))
  ds_dict["lon"] = (list(ds.lon.dims), da.from_array(ds.lon.values, chunks=(-1, -1)))
  ds_dict["t"] = (list(ds.time.dims), da.from_array(ds.time.values, chunks=(10)))
  ds_chunks = xr.Dataset(ds_dict)
  ```
  and this is not optimal. Either we find a way to make this chunking more flexible (using the `chunking` key from the config file?) or if there is a smarter way to do this. I have tested this approach for 1 year of DANRA data and it works with the time chunks of 10 works. This is a comparison of the timings (based on the log output) when adding the TOA radiation as forcing compared to not adding it:
  - **With** TOA radiation: 1 min 23 s - 54 GB dataset (30 min 37 s on my own laptop)
  - **Without** TOA radiation: 1 min 13 s - 43 GB dataset (26 min 47 s on my own laptop)

[...]

PR #30, opened 2024-10-30
move input coordinates to extra subsection

observingClouds:
- introduce `coordinates` subsection
- prepare for additional variable subsections (e.g. `dependencies` as discussed in #29 )


leifdenby:
> * introduce `coordinates` subsection

I think this is a good idea, but we should discuss how to deprecate/change the config schema. 

> [ ] Where are the config schemas defined? The tests contained both version `0.1.0` and `0.2.0`

We should discuss this. Sorry this is missing in the README, I could've typed up my current thinking.

I don't have a clear plan for this though, so would appreciate your input. I think it is quite hard to reconcile having a config schema version and a package version. I had thought that instead of versioning the schema separately from the package it might be simpler do to this together. My idea was that every time we create a new schema that would be versioned with the same version number as the package. As in, if we see the need to change the schema then that would only be allowed to change when we release a new tagged version of the package, and then schema version will match that same package version. This answer the question of "with this given config that uses a the config spec version what *minimum* version of mllam-data-prep will I need?", but doesn't handle backward compatibility. It might be that we should enforce that configs would be backwards compatible, but for how long?

observingClouds:
I feel like the versioning and deprecating of configs should be something that is well established, but I could not really find any general solutions.

I currently think we should:
- collect the schemas in a separate folder for better visibility and being able to keep several versions at the same time
- use pydantic to validate input and define migration functions and version incompatibilities

joeloskarsson:
I like the idea of the schema version matching the package version, keeps things simple. 

For now I don't think there is a need to worry about backwards compatibility until a 1.0.0 release. Better to build things to a state where it is specified in a way that we find useful to work with. Then that version can be fixed. After 1.0.0 one could follow semantic versioning for the schema as well, only allowing breaking backwards compat. when releasing new major versions.

leifdenby:
just a quick comment on your implementation here. Currently you are suggesting to put selection along coordinates into a new subsection called `coordinates` in the config for selecting a specific variable 

Screenshot 2024-11-01 at 20 37 02
(sorry I'm not sure how to write diff comment...)

I think is sensible since this is more explicit, but maybe we should call the subsection `sel` (to indicate that it uses the `xr.DataArray.sel(...)` method) or `sel_coords` or something like that? `coords` might suggest we are somehow *setting* the coordinates instead. I had initially not introduced this subsection because I couldn't think of anything else we might want to do. But one might for example want to ensure that the units are correct for a the variable you are selecting. In general I am realising now that it is currently implicit that the `variables` section is about *selecting* variables, whereas one might mistakingly assume that it is also possible to _set_ things (like the units if they are missing on a variable). I can see a need for that in future if the source file someone is trying to read doesn't have for example the `long_name` or `units` attributes set. What do you think?

observingClouds:
That this section is mostly about selecting has indeed be not clear to me. However, if we call the subsection `coordinates` both the selection as well as the expected units can be part of this section. If we would use `sel` or `sel_coords`, `units` would no longer fit to this section.

Issue #33, opened 2024-11-01
Make `mllam-data-prep` aware of geographical coordinates (by introducing projection info)


leifdenby:
# The problem

Currently, `mllam-data-prep` has no understanding of projection information, which means that it doesn't know how to define the latitude and longitude coordinate values for a given input dataset or how to define the same in zarr datasets produced by `mllam-data-prep`. We could of course *require* that latitude and longitude values are given for example as variables with the specific names (`lat` and `lon` for example) in each input dataset, but that seems limiting to me since deriving latitude and longitude coordinate values from the coordinate values actually used in a dataset is exactly what the projection information is for. If a given input dataset is in fact given on a lat/lon grid I vote we make that explicit (by requiring that the projection be set as `cartopy.crs.PlateCarree`) - give example configs below.

To my mind being able to define the latitude and longitude coordinate values is necessary for:

1. in `neural-lam` these are required so that `lat` / `lon` can be returned by a datastore and the projection used when plotting
2. being able to define the lat/lon coordinate convex hull that @TomasLandelius will use in implementing the geographical coordinate polygon-based approach to combining datasets on different spatial domains (e.g. DANRA and ERA5), https://github.com/mllam/weather-model-graphs/issues/30

# Proposed solution

I propose to provide two ways of deriving projection information in `mllam-data-prep`:

1. Implement functionality to parse the projection information from input datasets where projection information is stored in a [CF-compliant manner](http://cfconventions.org/cf-conventions/cf-conventions.html#grid-mappings-and-projections).
2. Introduce a new `projection` information config parameter available on `InputDataset`'s in the config, where the type of projection (given as a `cartopy.crs.Projection` `ClassName`, e.g. `LambertConformal`)

The new `projection` section to the `input` section would have three values to set the dimension names for the dimensions used in the projection (called `dims`), the `cartopy.crs.Projection`-derived projection class name (`class_name`) and the kwargs passed to this `ProjectionClass` to create an instance of the projection object. In the case where we can parse the projection from from CF-compliant attributes and the user has also defined projection in the config we warn the user that they are overwriting the projection information.

I will give two examples of what this would look like:

a) For DANRA (which is given on a Lambert Conformal grid) this would look like the following:

```yaml
inputs:
  danra_surface:
    path: https://mllam-test-data.s3.eu-north-1.amazonaws.com/single_levels.zarr
    projection:
      dims: [x, y]
      class_name: LambertConformal
      kwargs:
        central_longitude: 15.0
[...]

PR #34, opened 2024-11-06
Add ability to derive variables and add selected derived forcings

ealerskans:
Implements the ability to derive fields from the input datasets, as discussed in [Deriving forcings #29](https://github.com/mllam/mllam-data-prep/issues/29).

At the moment, I have only added the possibility to derive the following forcings:
- [x] top-of-atmosphere radiation
- [x] hour of day (cyclically encoded)
- [x] day of year (cyclically encoded)
- [ ] ~~time of year (cyclically encoded)~~

But additional variables, such as boundary and land-sea masks, should be added. But I think that is for another PR.


ealerskans:
@leifdenby I think this is ready for review now. However, I have one question, which I think could be a review point, and that is that the datetime forcings (`hour_of_day` and `day_of_year`) only have dimensions `(time)`. This means that it is not possible at the moment to only have this derived forcing (although it seems to work when I also derive `toa_radiation`). Should the dimensions of these datetime forcings perhaps be extended to `grid_index`/`(x,y)` as well or what would be the best way to do this?

mllam_data_prep/config.py
```diff
@@ -151,13 +169,20 @@ class InputDataset:
...
+        takes `time`, `lat` and `lon` as arguments.
     """
 
     path: str
     dims: List[str]
-    variables: Union[List[str], Dict[str, Dict[str, ValueSelection]]]
     dim_mapping: Dict[str, DimMapping]
     target_output_variable: str
+    variables: Union[List[str], Dict[str, Dict[str, ValueSelection]]] = None
+    derived_variables: Dict[str, DerivedVariable] = None
```
ealerskans, line 185: Since we now can have either `variables` or `derived_variables`, `variables` can no longer be required and has to be optional. I couldn't figure out a way to specify that we need to require either `variables`  or `derived_variables` and that both cannot be missing. Any ideas?

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    for _, derived_variable in derived_variables.items():
+        required_kwargs = derived_variable.kwargs
+        function_name = derived_variable.function
+        derived_variable_attributes = derived_variable.attributes or {}
+
+        # Separate the lat,lon from the required variables as these will be derived separately
+        latlon_coords_to_include = {}
+        for k, v in list(required_kwargs.items()):
+            if k in ["lat", "lon"]:
+                latlon_coords_to_include[k] = required_kwargs.pop(k)
```
ealerskans, line 49: This is so that we can remove the `lat` and `lon` variables since they should be retrieved based on the projection information and not directly taken from the input dataset in the future (#33). However, this is not a very general formulation (since I write `if k in ["lat", "lon"]`) since this requires that they are named `lat` and `lon` in the source dataset and not something else (like `latitude` and `longitude`). But I couldn't come up with something more general. Any ideas?

ealerskans:
I realized that I haven't added any tests for this update so I am looking into that. I have very, very little experience with writing tests (basically none) but I will try and add something. So if you have any suggestions/comments/opinions on what to test (and how) then I am all ears!

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+                )
+                field.attrs[attribute] = field_attributes[attribute]
+            else:
+                # Attributes are set and nothing has been defined in the config file
+                pass
+
+    return field
+
+
+def _return_dropped_coordinates(ds_subset, ds_input, required_coordinates, chunks):
```
mafdmi, line 288: Add arguments and returns to docstring

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+        Dictionary with the variables to derive
+        with keys as the variable names and values with entries for
+        kwargs and function to use in the calculation
+    chunking: dict
+        Dictionary with keys as the dimensions to chunk along and values
+        with the chunk size
+
+    Returns
+    -------
+    ds : xr.Dataset
```
mafdmi, line 28: Skip name of return variable, since only the type matters

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+        function_name = derived_variable.function
+        derived_variable_attributes = derived_variable.attributes or {}
+
+        # Separate the lat,lon from the required variables as these will be derived separately
+        latlon_coords_to_include = {}
+        for k, v in list(required_kwargs.items()):
+            if k in ["lat", "lon"]:
+                latlon_coords_to_include[k] = required_kwargs.pop(k)
+
+        # Subset the dataset
```
mafdmi, line 51: It is a bit counterintuitive, that the ds_subset variable is not used when subsetting ds.. Maybe "Get input dataset for  calculating derived variables"

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+
+    Returns
+    -------
+    toa_radiation: xr.DataArray or float
+        TOA radiation data
+    """
+    logger.info("Calculating top-of-atmosphere radiation")
+
+    # Solar constant
+    E0 = 1366  # W*m**-2
```
mafdmi, line 318: User longer var name

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+
+    day = time.dt.dayofyear
+    hr_utc = time.dt.hour
```
mafdmi, line 321: I prefer "hour_utc", but not very important.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+
+    # Eq. 1.6.1a in Solar Engineering of Thermal Processes 4th ed.
+    dec = np.pi / 180 * 23.45 * np.sin(2 * np.pi * (284 + day) / 365)
+
+    hr_lst = hr_utc + lon / 15
```
mafdmi, line 326: Not sure what this variable is, based on the variable name

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    hr_angle = 15 * (hr_lst - 12)
```
mafdmi, line 327: hour_angle

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+
+    # Eq. 1.6.2 with beta=0 in Solar Engineering of Thermal Processes 4th ed.
+    cos_sza = np.sin(lat * np.pi / 180) * np.sin(dec) + np.cos(
```
mafdmi, line 330: Can cos_sza be defined in words in a comment or will it be too long?

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    logger.info("Calculating top-of-atmosphere radiation")
+
+    # Solar constant
+    E0 = 1366  # W*m**-2
+
+    day = time.dt.dayofyear
+    hr_utc = time.dt.hour
+
+    # Eq. 1.6.1a in Solar Engineering of Thermal Processes 4th ed.
+    dec = np.pi / 180 * 23.45 * np.sin(2 * np.pi * (284 + day) / 365)
```
mafdmi, line 324: Can dec be defined in words in a comment or will it be too long?

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    # Cyclic encoding of hour of day
+    hour_of_day_cos, hour_of_day_sin = cyclic_encoding(hour_of_day, 24)
+
+    if isinstance(hour_of_day_cos, xr.DataArray):
+        # Add attributes
+        hour_of_day_cos.name = "hour_of_day_cos"
+        hour_of_day_cos.attrs[
+            "long_name"
+        ] = "Cosine component of cyclically encoded hour of day"
+        hour_of_day_cos.attrs["units"] = "1"
```
mafdmi, line 376: "1" is used to indicate no unit? Would it be better with None or empty string?

mafdmi:
> @leifdenby I think this is ready for review now. However, I have one question, which I think could be a review point, and that is that the datetime forcings (`hour_of_day` and `day_of_year`) only have dimensions `(time)`. This means that it is not possible at the moment to only have this derived forcing (although it seems to work when I also derive `toa_radiation`). Should the dimensions of these datetime forcings perhaps be extended to `grid_index`/`(x,y)` as well or what would be the best way to do this?

I talked with @leifdenby about this, and he said that the data is compressed anyway, when saving to zarr, so it shouldn't increase the size of the dataset that much. Let's test it maybe?

mllam_data_prep/config.py
```diff
@@ -151,13 +169,20 @@ class InputDataset:
...
     attributes: Dict[str, Any] = None
```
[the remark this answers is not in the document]
ealerskans: I am not sure that I understand what isn't documented. Do you mean that `variables` and `derived_variables` are not documented in the docstring to `InputDataset` or something else? Because the docstring does include documentation of `variables` and `derived_variables` (lines 155-159 + 172-173; sorry not sure how to format this with the lines highlighted in a reply to a comment).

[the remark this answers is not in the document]
mafdmi: No, it was just, there is no entry in the docstring for the "attributes" attribute ;)

mllam_data_prep/derived_variables.py
ealerskans: According to [CF conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03.html)
> The conforming unit for quantities that represent fractions, or parts of a whole, is "1"

I would guess that this applies for sine and cosine quantities?

mllam_data_prep/config.py
```diff
@@ -120,7 +137,8 @@ class InputDataset:
         1) the path to the dataset,
         2) the expected dimensions of the dataset,
         3) the variables to select from the dataset (and optionally subsection
-           along the coordinates for each variable) and finally
+           along the coordinates for each variable) and/or the variables to derive
```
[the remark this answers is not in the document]
ealerskans: Hmm I haven't really thought about if they should be mutually exclusive or if we should be able to have both in the same named dataset under `inputs`. But at the moment we can't since I have not added any feature of merging the subset from `variables` with the subset from `derived_variables`.

If we want to have them mutually exclusive then `mllam_data_prep/config.py` would need to reflect that as well, as you have observed @mafdmi. Then in `mllam_data_prep/derived_variables.py` line 154 should be an `elif` and not an `if` I guess,:
```diff
- if derived_variables:
+ elif derived_variables:
```

If we don't want to have them mutually exclusive, we could perhaps try and merge them after having derived them. But this will then require some additional development.

Thoughts @leifdenby ?

leifdenby:
> @leifdenby I think this is ready for review now. However, I have one question, which I think could be a review point, and that is that the datetime forcings (`hour_of_day` and `day_of_year`) only have dimensions `(time)`. This means that it is not possible at the moment to only have this derived forcing (although it seems to work when I also derive `toa_radiation`). Should the dimensions of these datetime forcings perhaps be extended to `grid_index`/`(x,y)` as well or what would be the best way to do this?

In the case where a derived field is only a function of say time, I think it might be best to simply broadcast these so that the values are duplicated across all gridpoints, that way making sure the dimension are still say `[time, x, y]`. What you've brought up here is an interesting point though: what happens if you select variables from a single input dataset that have a different set of dimensions? I would expect that the code currently would broadcast the missing coordinates before stacking all variables along the feature vector, but I am not sure to be honest... I will give what you written a try! :)

mllam_data_prep/derived_variables.py
leifdenby: "1" is common for dimensionless variables. In fact this doesn't mean "does not have a unit" but the unit is dimensionless :) We should keep "1" since that that is also the CF compliant way of indicating that a variable is dimensionless (http://cfconventions.org/cf-conventions/cf-conventions.html#dimensionless-units)

example.danra.yaml
```diff
@@ -73,6 +73,32 @@ inputs:
...
+          lat: lat
+          lon: lon
+        function: mllam_data_prep.derived_variables.calculate_toa_radiation
+      hour_of_day:
+        kwargs:
+          time: time
+        function: mllam_data_prep.derived_variables.calculate_hour_of_day
+    dim_mapping:
+      time:
+        method: rename
```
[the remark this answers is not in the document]
leifdenby: In principle we don't since they have the same name, but as it is currently implemented all the dimensions in the output must be given in the `dim_mapping` part of the config, and there isn't an option for "this name of this dimension is unchanged". You are welcome to add that to the readme if you like :) I just did it like this to be explicit rather than implicit about what is going on.

leifdenby:
> @leifdenby Should we consider adding a type check to the pre-commit and encourage people to write type hints?

We could do, but I think we should ask people who have contributed to the code first. If we want to look for type hints in pre-commit then we should enforce that they are present, but it would be good to check that everyone is happy with that before we go ahead and make that change I think.

mllam_data_prep/config.py
leifdenby: That is a good point. I think it is fine to make them both optional here in the spec and then we can enforce this when we load the config :) So probably we should create a `load_config()` function that for now wraps `Config.from_yaml_file()` (https://github.com/ealerskans/mllam-data-prep/blob/feature/derive_forcings/mllam_data_prep/create_dataset.py#L286). This kind of check can't be done at the type level (which is what dataclasses prescribe) but one would have to use something like [pydantic](https://docs.pydantic.dev/latest/) for that. I haven't added use of pydantic yet because until now I didn't think we needed it and I wanted to add yet another dependency. But we could add that in a separate PR. For now I would just wrap the `Config.from_yaml_file()` call, you could maybe include an inline comment to say that this check could eventually be done with pydantic.

mllam_data_prep/derived_variables.py
leifdenby: I think this is fine for now :) We can wait and not merge your work here until #33 is complete, or we can merge your's first and then adapt with #33. I know @observingClouds is on holiday now, so maybe for now simply issue a warning to say that you are assuming that the lat/lon coordinates are given as variables called `lat` and `lon`

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    return day_of_year_cos, day_of_year_sin
+
+
+def cyclic_encoding(data_array, da_max):
+    """
+    Cyclic encoding of data
+
+    Parameters
+    ----------
+    da : xr.DataArray
```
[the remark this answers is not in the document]
leifdenby: `da` is common prefix for `xr.DataArray` variables when working with xarray, as is `ds` for `xr.Dataset`s. This is also what you find elsewhere in the codebase :)

leifdenby:
I will continue my review tomorrow, looking good so far!

mafdmi:
Just got a point from @leifdenby, that my review comments could be understood in a very "do this and that" manner, or that my comments can be understood as I think something is good or bad written.. 
So I just want to make it clear, that all my comments are only suggestions, and it's totally fine to say, "hm I think what I wrote originally is more appropriate". My focus as a reviewer is to comment where I see, that something might be difficult to understand/read, or if I see something that I think could be written in a more clear way in sake of the future people having to read/understand the code. This is of coarse subjective, so feel free to disagree with my comments!:) Also, I really focus on the code when I review - I really don't have an intension to judge anybody's work!
I could add more "maybe do this", "consider do this", "it would be good to..." - I just find it a bit long sometimes, and since I see the review only as suggestions for me sentences like these are often implicit in my comments.
But I can also see, that it can be interpreted differently if not stating clearly, that they are suggestions, so I will do this more from now on:)

mafdmi:
> > @leifdenby Should we consider adding a type check to the pre-commit and encourage people to write type hints?
> 
> We could do, but I think we should ask people who have contributed to the code first. If we want to look for type hints in pre-commit then we should enforce that they are present, but it would be good to check that everyone is happy with that before we go ahead and make that change I think.
> 
> It is probably best to make that a separate issue and then tag people to get some feedback. Since that is a separate issue to the one being worked on here :)

Agree:)

ealerskans:
> Just got a point from @leifdenby, that my review comments could be understood in a very "do this and that" manner, or that my comments can be understood as I think something is good or bad written.. So I just want to make it clear, that all my comments are only suggestions, and it's totally fine to say, "hm I think what I wrote originally is more appropriate". My focus as a reviewer is to comment where I see, that something might be difficult to understand/read, or if I see something that I think could be written in a more clear way in sake of the future people having to read/understand the code. This is of coarse subjective, so feel free to disagree with my comments!:) Also, I really focus on the code when I review - I really don't have an intension to judge anybody's work! I could add more "maybe do this", "consider do this", "it would be good to..." - I just find it a bit long sometimes, and since I see the review only as suggestions for me sentences like these are often implicit in my comments. But I can also see, that it can be interpreted differently if not stating clearly, that they are suggestions, so I will do this more from now on:)

No worries at all! I didn't take it as you judging my work at all. I really appreciate the review and suggestions (I didn't take them as anything else). I see it as a way to improve what I write, to make it more clear and understandable (and also to learn new ways to code in!) :D

mllam_data_prep/config.py
```diff
@@ -52,6 +52,23 @@ class ValueSelection:
...
+    are specified.
+
+    Attributes:
+        kwargs: Variables required for calculating the derived variable.
+        function: Function used to calculate the derived variable.
+    """
+
+    kwargs: Dict[str, str]
+    function: str
+    attributes: Dict[str, Any] = None
```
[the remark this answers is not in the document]
mafdmi: With "bad" I just mean "inappropriate", since the "attributes" attribute is itself.. and attribute, so it might be confusing to read.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+            if req_coord in chunks:
+                ds_input = ds_input.reset_coords(req_coord)
+
+        # Chunk the dataset
+        ds_input = _chunk_dataset(ds_input, chunks)
+
+        # Add function arguments to kwargs
+        kwargs = {}
+        if len(latlon_coords_to_include):
+            latlon = get_latlon_coords_for_input(ds)
```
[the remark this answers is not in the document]
ealerskans: The dummy function is preparation for #38, in which it will be possible to derive the `lat` and `lon` coordinates from projection information if they are not included in the dataset.

mafdmi:
> > Just got a point from @leifdenby, that my review comments could be understood in a very "do this and that" manner, or that my comments can be understood as I think something is good or bad written.. So I just want to make it clear, that all my comments are only suggestions, and it's totally fine to say, "hm I think what I wrote originally is more appropriate". My focus as a reviewer is to comment where I see, that something might be difficult to understand/read, or if I see something that I think could be written in a more clear way in sake of the future people having to read/understand the code. This is of coarse subjective, so feel free to disagree with my comments!:) Also, I really focus on the code when I review - I really don't have an intension to judge anybody's work! I could add more "maybe do this", "consider do this", "it would be good to..." - I just find it a bit long sometimes, and since I see the review only as suggestions for me sentences like these are often implicit in my comments. But I can also see, that it can be interpreted differently if not stating clearly, that they are suggestions, so I will do this more from now on:)
> 
> No worries at all! I didn't take it as you judging my work at all. I really appreciate the review and suggestions (I didn't take them as anything else). I see it as a way to improve what I write, to make it more clear and understandable (and also to learn new ways to code in!) :D

Good to here! Just moderated my comments a bit anyway;) Good, and me too, as a reviewer, I also see the review process as a way to learn more and get wiser!

mllam_data_prep/derived_variables.py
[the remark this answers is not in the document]
ealerskans: Yes, but I had named the variable`data_array` in the code but `da` in the docstring, so I was not being consistent.
I have changed the names to `data` instead (will push the new commits soon) since `da` is what is usually used when importing `dask. array` as well :)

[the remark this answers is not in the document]
mafdmi: Yes, that was just what I meant:)

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+                    " This attribute has not been defined in the 'attributes' section"
+                    " of the config file either. Make sure that you add it to the"
+                    f" 'attributes' section of the derived variable '{field.name}'."
+                )
+        else:
+            if attribute in field_attributes.keys():
+                logger.warning(
+                    f"The attribute '{attribute}' of the derived field"
+                    f" {field.name} is being overwritten from"
+                    f" '{field.attrs[attribute]}' to"
```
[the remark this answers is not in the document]
ealerskans: It actually doesn't :)
I have tested all 4 combinations manually and they work as expected at the moment. But this is one of the things I would like to add a test for to make sure that it works as expected!

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    derived_field: xr.DataArray or tuple
+        The derived field
+    """
+    if isinstance(derived_field, xr.DataArray):
+        derived_field = _check_attributes(derived_field, derived_field_attributes)
+    elif isinstance(derived_field, tuple) and all(
+        isinstance(field, xr.DataArray) for field in derived_field
+    ):
+        for field in derived_field:
+            field = _check_attributes(field, derived_field_attributes)
```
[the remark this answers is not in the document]
ealerskans: Yes, it is :)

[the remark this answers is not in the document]
mafdmi: Nice:)

mllam_data_prep/config.py
```diff
@@ -52,6 +52,23 @@ class ValueSelection:
...
+@dataclass
+class DerivedVariable:
+    """
+    Defines a derived variables, where the kwargs (variables required
+    for the calculation) and the function (for calculating the variable)
+    are specified.
+
+    Attributes:
+        kwargs: Variables required for calculating the derived variable.
+        function: Function used to calculate the derived variable.
```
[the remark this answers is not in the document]
ealerskans: Done :)

mllam_data_prep/create_dataset.py
```diff
@@ -120,28 +121,56 @@ def create_dataset(config: Config):
 
     output_config = config.output
     output_coord_ranges = output_config.coord_ranges
+    chunking_config = config.output.chunking or {}
```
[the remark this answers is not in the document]
ealerskans: Good idea, I have done that now!

mllam_data_prep/create_dataset.py
```diff
@@ -120,28 +121,56 @@ def create_dataset(config: Config):
...
 
     dataarrays_by_target = defaultdict(list)
 
     for dataset_name, input_config in config.inputs.items():
         path = input_config.path
-        variables = input_config.variables
+        variables = input_config.variables or None
```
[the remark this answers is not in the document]
ealerskans: You are right, it is not needed.

mllam_data_prep/ops/loading.py
```diff
@@ -54,4 +57,8 @@ def load_and_subset_dataset(fp, variables):
             ) from ex
     else:
         raise ValueError("The `variables` argument should be a list or a dictionary")
+
+    chunks = {d: chunking.get(d, int(ds_subset[d].count())) for d in ds_subset.dims}
```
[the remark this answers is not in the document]
ealerskans: Thanks, I have updated various places according to this :)

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    -------
+    function: object
+        Function for deriving the specified variable
+    """
+    # Get the name of the calling module
+    calling_module = globals()["__name__"]
+
+    if "." in function_namespace:
+        # If the function name is a full namespace, get module and function names
+        module_name, function_name = function_namespace.rsplit(".", 1)
```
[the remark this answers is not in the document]
ealerskans: Not sure that it simplified things but I think it might be cleaner.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+        # Add the derived field(s) to the subset
+        if isinstance(derived_field, xr.DataArray):
+            ds_subset[derived_field.name] = derived_field
+        elif isinstance(derived_field, tuple) and all(
+            isinstance(field, xr.DataArray) for field in derived_field
+        ):
+            for field in derived_field:
+                ds_subset[field.name] = field
+        else:
+            raise TypeError(
```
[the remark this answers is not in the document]
ealerskans: This actually made things more simple! By moving the type check up I could remove the `_check_field` function and just use the `_check attributes` function directly instead.

mllam_data_prep/derived_variables.py
ealerskans: Done :)

mllam_data_prep/derived_variables.py
ealerskans: I agree. So I renamed the dataset `ds_subset` to `ds_derived_vars` and updated the comment :)

mllam_data_prep/derived_variables.py
ealerskans: Thanks for the comments on the `calculate_toa_radiation` function. I looked it through more carefully and also learned more what the different variables represent. I have tried to update the function so that it should be easier to understand it :)

mllam_data_prep/derived_variables.py
ealerskans: Done :)

mllam_data_prep/derived_variables.py
ealerskans: Done :)

mllam_data_prep/config.py
ealerskans: It sounds like a good idea, but I don't really know how it should be done :)

mllam_data_prep/derived_variables.py
ealerskans: Sounds good to me. I have added a warning for this.

mllam_data_prep/config.py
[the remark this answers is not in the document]
ealerskans: It is only overwriting if the attributes have been set in both the config file and the function used to derive the variable so `overwrite_attributes`  is not always appropriate either. I don't have any strong feelings if we're using `attributes`  or `da_attribtes` :)

ealerskans:
> Consider to generalize this into a load_dataset function and re-used in the `load_and_subset_dataset` function - maybe not very important.

I think this is a good idea and have actually done that. So I have added a `load_dataset`function and now the dataset is loaded outside both `load_and_subset_dataset` (which I have renamed to `subset_dataset`) and `derived_variables`.

However, I have not used error handling a lot so I just added the Exception raises to `load_dataset`, `subset_dataset` and `derive_variables`. It looks kind of messy now I think... So I would really like some feedback/suggestion for how to do this instead :)

mllam_data_prep/config.py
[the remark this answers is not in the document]
ealerskans: Ah, I see. This was added before this PR but I have made an attempt to document it in the docstring. Hopefully it is correct, @leifdenby ?

mllam_data_prep/config.py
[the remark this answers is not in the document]
ealerskans: I have removed the 'and'.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,478 @@
...
+    -------
+    function: object
+        Function for deriving the specified variable
+    """
+    # Get the name of the calling module
+    calling_module = globals()["__name__"]
+
+    # Get module and function names
+    function_namespace_list = function_namespace.rsplit(".")
+    if len(function_namespace_list) > 1:
```
mafdmi, line 176: You can actually avoid this whole if-statement if you do
`module_name, _, function_name = function_namespace.rpartition(".")

mafdmi:
> > Consider to generalize this into a load_dataset function and re-used in the `load_and_subset_dataset` function - maybe not very important.
> 
> I think this is a good idea and have actually done that. So I have added a `load_dataset`function and now the dataset is loaded outside both `load_and_subset_dataset` (which I have renamed to `subset_dataset`) and `derived_variables`.
> 
> However, I have not used error handling a lot so I just added the Exception raises to `load_dataset`, `subset_dataset` and `derive_variables`. It looks kind of messy now I think... So I would really like some feedback/suggestion for how to do this instead :)

Very nice with the `load_dataset` function. Yes, agree it becomes a bit messy with all the exceptions. I'll wait to see what @leifdenby says about the reason for having them.

mllam_data_prep/config.py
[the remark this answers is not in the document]
mafdmi: Maybe attributes is clear enough, if just people read the readme, which I guess we can assume they do.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+
+    try:
+        ds = xr.open_zarr(fp)
+    except ValueError:
+        ds = xr.open_dataset(fp)
+
+    ds_subset = xr.Dataset()
+    ds_subset.attrs.update(ds.attrs)
+    for _, derived_variable in derived_variables.items():
+        required_kwargs = derived_variable.kwargs
```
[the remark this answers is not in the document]
mafdmi: I think it's fine to keep it as is, since I don't think my suggestion makes it simpler, and since the README states clearly what the kwargs section is used to.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+        Dictionary with keys as dimensions to be chunked and
+        chunk sizes as the values
+
+    Returns
+    -------
+    ds: xr.Dataset
+        Dataset with chunking applied
+    """
+    # Define the memory limit check
+    memory_limit_check = 1 * 1024**3  # 1 GB
```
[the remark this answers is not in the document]
leifdenby: Yes. I would put this constant into a global variable, something like `CHUNK_MAX_SIZE_WARNING`, but I would also put this "chunk size check into its own function somewhere outside of `ops`, maybe `mllam_data_prep.chunking`, or maybe `mllam_data_prep.ops.chunking` makes sense? I think this is something we might want to reuse :)

mllam_data_prep/derived_variables.py
leifdenby: Rather than using `globals()` which would require that the module is already imported I think it might be better to use the `importlib` package.

I think you would write:

```python
module = importlib.import_module(module_name)
fn = getattr(module, function_name)
```

mllam_data_prep/config.py
ealerskans: Thanks, I will try to implement this!

leifdenby (commented):
I've made a few more suggestions. Hope it makes sense. Looking good! 🚀

mllam_data_prep/config.py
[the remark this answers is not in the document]
ealerskans: > But in the case for derived variables (which would be a different dataclass, the `DerivedVariable` one I thought we had said that we wanted to provide the user a way to _set the attributes in the case where what is returned from a function doesn't have those required attributes_. Am I remembering that correctly?

No, you are correct. The `attributes` for the `DerivedVariable` class is a way to set attributes for the derived variables. Form what I remember, we discussed that it should do the following (which it also does at the moment):
- If the attributes (at the moment `units` and `long_name`) are not set by the function then they need to be specified in the config file through the `attributes` section of the `DerivedVariable` 
- If the attributes are set in the function and also defined in the config file then the attributes from the config file will overwrite the ones from the function and a warning will be issued.
- If the attributes are set in the function and not defined in the config file, nothing will happen. 
- If the attributes are not set in the function nor in the config file, we raise an error

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,478 @@
...
+        # Calculate the derived variable
+        derived_field = func(**kwargs)
+
+        # Check the derived field(s) and add it to the dataset
+        if isinstance(derived_field, xr.DataArray):
+            derived_field = _check_attributes(
+                derived_field, derived_variable_attributes
+            )
+            ds_derived_vars[derived_field.name] = derived_field
+        elif isinstance(derived_field, tuple) and all(
```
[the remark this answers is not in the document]
ealerskans: The functions that returns two data-arrays currently are the ones where we drive day of year and hour of day. Since I have implemented a cyclic encoding in them, the cosine and sine part are both returned. I thought that it wouldn't make any sense to return only e.g. hour of day as a non-cyclically encoded variable and that is why I did it like this. Not sure if we wanted to split them up how that should be implemented then. The other option is then to not do the cyclic encoding here.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,478 @@
...
+        day_of_year_sin.name = "day_of_year_sin"
+        day_of_year_sin.attrs[
+            "long_name"
+        ] = "Sine component of cyclically encoded day of year"
+        day_of_year_sin.attrs["units"] = "1"
+
+    return day_of_year_cos, day_of_year_sin
+
+
+def cyclic_encoding(data, data_max):
```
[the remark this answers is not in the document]
ealerskans: I agree that it is getting long and it would be nice to split it up. The main reason for not adding it to `mllam_data_prep.ops` is that I don't know what "ops" stands for and what can/should be included here.

The idea and structure makes sense but the implementation is not very clear, probably because I am not very familiar with object oriented programming.

mllam_data_prep/ops/loading.py
```diff
@@ -1,27 +1,48 @@
...
     except ValueError:
         ds = xr.open_dataset(fp)
 
+    return ds
+
+
+def subset_dataset(ds, variables, chunking):
+    """
+    Load the dataset, subset the variables along the specified coordinates and
+    check coordinate units
```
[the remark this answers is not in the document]
ealerskans: Yes, I agree :)

mllam_data_prep/derived_variables.py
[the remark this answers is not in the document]
mafdmi: I have also been a bit puzzeled about what the "ops" stands for. I think I have interpreted is as "operations", but not sure if this is correct. Do we document the meaning somewhere? Otherwise, we could maybe consider to rename it to make it more self-explainable (for another PR then).

mllam_data_prep/config.py
```diff
@@ -52,6 +52,24 @@ class ValueSelection:
...
+        attributes: Attributes (e.g. `units` and `long_name`) for the derived variable.
```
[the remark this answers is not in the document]
ealerskans: I don't have any strong feelings if the call it `attributes` or `attrs` but I think I prefer `attrs` (since it is shorter and still easily understood what it means). I have changed it now :)

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,478 @@
...
+        # Add function arguments to kwargs
+        kwargs = {}
+        if len(latlon_coords_to_include):
+            latlon = get_latlon_coords_for_input(ds)
+            for key, val in latlon_coords_to_include.items():
+                kwargs[val] = latlon[key]
+        kwargs.update({val: ds_input[key] for key, val in required_kwargs.items()})
+        func = _get_derived_variable_function(function_name)
+        # Calculate the derived variable
+        derived_field = func(**kwargs)
```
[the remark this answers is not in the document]
ealerskans: Currently I have the check after `derived_field = func(**kwargs)` where I check on if `field` is an `xr.DataArray` or `Tuple(xr.DataArray)` (since currently the time forcings are outputted as tuples since they have both cos and sin components).

I have, however, not added any functionality for turning `field` into an `xr.DataArray`. Should this be added perhaps?

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,478 @@
...
+
+        # Check the derived field(s) and add it to the dataset
```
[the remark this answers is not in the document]
ealerskans: Good suggestions. I have implemented them :)

I have also added to the README that `long_name` and `units`  are required attributes.

mllam_data_prep/create_dataset.py
```diff
@@ -120,28 +121,56 @@ def create_dataset(config: Config):
...
-            expected_attributes=expected_input_attributes,
-            dataset_name=dataset_name,
-        )
+        if variables:
+            logger.info(f"Loading dataset {dataset_name} from {path} and subsetting")
+            try:
+                ds = load_and_subset_dataset(
+                    fp=path, variables=variables, chunking=chunking_config
+                )
+            except Exception as ex:
```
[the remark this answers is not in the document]
ealerskans: I have now removed the exception wrapping for `subset_dataset` and `derive_variables`. But I have kept it wrapped around the `load_input_dataset` function call since I didn't want to have to pass both `dataset_name` and `path` to the `load_input_dataset` function. Let me know what you think.

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
...
+    """
+    logger.info("Deriving variables")
+
+    try:
+        ds = xr.open_zarr(fp)
+    except ValueError:
+        ds = xr.open_dataset(fp)
+
+    ds_subset = xr.Dataset()
+    ds_subset.attrs.update(ds.attrs)
```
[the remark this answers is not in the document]
ealerskans: I have updated the code with the suggestions:
- split out the "load" part into its own function (`load_input_dataset`)
- moved the "subsetting" part to its own function (`susbet_dataset`) and moved it to its own module (`ops.subsetting`)
- renamed the return from `load_input_dataset` to `ds_input`
- call `load_input_dataset` from `create_dataset`  and pass `ds_input` to both `subset_dataset` and `derive_variables`

mllam_data_prep/derived_variables.py
```diff
@@ -0,0 +1,459 @@
+import importlib
```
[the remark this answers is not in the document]
ealerskans: Done!

mllam_data_prep/config.py
```diff
@@ -151,13 +169,20 @@ class InputDataset:
         (e.g. two datasets that coincide in space and time will only differ in the feature dimension,
         so the two will be combined by concatenating along the feature dimension).
         If a single shared coordinate cannot be found then an exception will be raised.
+    derived_variables: Dict[str, DerivedVariable]
+        Dictionary of variables to derive from the dataset, where the keys are the variable names and
```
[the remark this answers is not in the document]
ealerskans: Agree. I have updated with Leif's suggestion.

mllam_data_prep/config.py
```diff
@@ -151,13 +169,20 @@ class InputDataset:
...
+        the values are dictionaries defining the necessary function and kwargs. E.g.
+        `{"toa_radiation": {"kwargs": {"time": "time", "lat": "lat", "lon": "lon"}, "function": "calculate_toa_radiation"}}`
+        would derive the "toa_radiation" variable using the `calculate_toa_radiation` function, which
+        takes `time`, `lat` and `lon` as arguments.
```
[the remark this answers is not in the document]
ealerskans: True. I have updated with your suggestion :)

mllam_data_prep/derived_variables.py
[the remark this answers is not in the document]
mafdmi: I think that would be good to have too

mllam_data_prep/create_dataset.py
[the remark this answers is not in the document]
mafdmi: Looks good to me:)

mllam_data_prep/config.py
ealerskans: `variables` and `derived_variables` are no longer mutually exclusive - we can have both in the same dataset now :)
I have taken inspiration from your solution @mafdmi and added a check which requires us to have defined at least one of `variables` and `derived_variables` and also a check so that they both don't contain the same variable names.

README.md
```diff
@@ -187,6 +187,32 @@ inputs:
         name_format: "{var_name}"
     target_output_variable: forcing
 
+  danra_derived_forcings:
```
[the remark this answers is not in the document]
ealerskans: That wasn't possible before but i have now made it so that we can add both `variables` and `derived_variables` from the same dataset so I have moved everything under the other input group, as you suggested :)

mllam_data_prep/create_dataset.py
```diff
@@ -120,23 +125,50 @@ def create_dataset(config: Config):
...
+
+        # Initialize the output dataset and add dimensions
+        ds = xr.Dataset()
+        ds.attrs.update(ds_input.attrs)
+        for dim in ds_input.dims:
+            ds = ds.assign_coords({dim: ds_input.coords[dim]})
+
+        if variables:
+            logger.info(f"Subsetting dataset {dataset_name}")
+            ds = subset_dataset(
```
leifdenby, line 156: I think it would better to keep to loop here, so that the `xr.Dataset` you are accumulating into is given here, rather than `ds` being modified inside `subset_dataset()`. Or said another way: it is easier to reason about what a function does if it doesn't modify its arguments but instead returns the result.

So I would write:

```python
for var_name in variables:
    ds[var_name] = extract_variable(ds=ds_input, var_name=var_name, ...)
```

mllam_data_prep/create_dataset.py
```diff
@@ -120,23 +125,50 @@ def create_dataset(config: Config):
...
+                ds_subset=ds,
+                ds_input=ds_input,
+                variables=variables,
+                chunking=chunking_config,
+            )
+
+        if derived_variables:
+            logger.info(f"Deriving variables from {dataset_name}")
+            ds = derive_variables(
```
leifdenby, line 165: Same comment as above :) It would be better to return one variable at a time and assigning the returned `xr.DataArray` into the dataset you have defined here.

mllam_data_prep/ops/derived_variables.py
```diff
@@ -0,0 +1,511 @@
...
+            derived_field = _check_for_required_attributes(
+                derived_field, expected_field_attributes
+            )
+            derived_field = _return_dropped_coordinates(
+                derived_field, ds_subset, required_coordinates, chunks
+            )
+            derived_field = _align_derived_variable(
+                derived_field, ds_input, target_dims
+            )
+            ds[derived_field.name] = derived_field
```
leifdenby, line 108: I have a similar comment here to the one where you call the function to create derived variables and subset a dataset to extract variables :)

Rather than modifying the argument inside the function it would be better here to only return the thing that has to change and assign the change here in the calling function. Also, I don't quite understand why `_check_for_required_attributes` needs to return anything? Something that "checks" just raises an exception or warning, no?

mllam_data_prep/ops/derived_variables.py
```diff
@@ -0,0 +1,511 @@
...
+        }
+        required_coordinates = [
+            req_var for req_var in required_kwargs.keys() if req_var in ds_subset.coords
+        ]
+        ds_subset = ds_subset.drop_indexes(required_coordinates, errors="ignore")
+        for req_coord in required_coordinates:
+            if req_coord in chunks:
+                ds_subset = ds_subset.reset_coords(req_coord)
+
+        # Chunk the dataset
```
leifdenby, line 81: it would be good to include a comment here on why we need to apply the chunking here :)

mllam_data_prep/ops/derived_variables.py
```diff
@@ -0,0 +1,511 @@
...
+            field = field.expand_dims({dim: ds.sizes[dim]})
+
+    # Broadcast to match only the target dimensions
+    broadcast_shape = {dim: ds[dim] for dim in target_dims if dim in ds.dims}
+    field = field.broadcast_like(xr.Dataset(coords=broadcast_shape))
+
+    return field
+
+
+def calculate_toa_radiation(lat, lon, time):
```
leifdenby, line 320: how do you feel about moving these functions that actually do the calculations to a separate module? This file is getting quite a long and that way when people want to add a new function for deriving a variable it will be simpler to work out where to add it :)

What I am suggesting is that you change the file/folder structure from:

```bash
ops/__init__.py
...
ops/derived_variables.py
```

to

```bash
ops/__init__.py
...
ops/derived_variables/__init__.py # in here you would just import the `derive_variable` function so that it can be imported by the parent module in the same way as you currently have it :)
ops/derived_variables/main.py # this could contain the `derive_variable` function and the others that check attributes etc, and would import the functions that actually to the calculations from the modules below:
ops/derived_variables/physical.py # this could be for the derived variables which have a "physical" nature, for example top-of-atmosphere radiation
ops/derived_variables/time_encoding.py # this could contain the time encoding functions

File: mllam_data_prep/ops/derived_variables.py [hunk 47]
```diff
@@ -0,0 +1,511 @@
...
+    # Eq. 1.6.2 with beta=0 in Solar Engineering of Thermal Processes 4th ed.
+    # cos_sza: Cosine of solar zenith angle
+    cos_sza = np.sin(lat * np.pi / 180) * np.sin(dec) + np.cos(
+        lat * np.pi / 180
+    ) * np.cos(dec) * np.cos(hour_angle * np.pi / 180)
+
+    # Where TOA radiation is negative, set to 0
+    toa_radiation = xr.where(solar_constant * cos_sza < 0, 0, solar_constant * cos_sza)
+
+    if isinstance(toa_radiation, xr.DataArray):
```
leifdenby, line 373: perfect!

mllam_data_prep/ops/derived_variables.py
```diff
@@ -0,0 +1,511 @@
...
+    Function for calculating hour of day features with a cyclic encoding
+
+    Parameters
+    ----------
+    time : Union[xr.DataArray, datetime.datetime]
+        Time
+
+    Returns
+    -------
+    hour_of_day_cos: Union[xr.DataArray, float]
```
leifdenby, line 393: It might be simpler (and more explicit which is often good) to change the config to explicitly call these features `hour_of_day_cos` and `hour_of_day_sin`, and that way each function only every returns one derived variable (which makes the rest of the code simpler I think)

mllam_data_prep/ops/derived_variables.py
```diff
@@ -0,0 +1,511 @@
...
+
+    Returns
+    -------
+    data_cos: Union[xr.DataArray, float, int]
+        Cosine part of cyclically encoded input data
+    data_sin: Union[xr.DataArray, float, int]
+        Sine part of cyclically encoded input data
+    """
+
+    data_sin = np.sin((data / data_max) * 2 * np.pi)
```
leifdenby, line 503: you could split this into two functions or make `cos` or `sin` (as strings) an argument to this function say `cyclically_encode_values(values, max_value, component='cos')`. But inline might be better to make this more explicit (since it is only one line)

mllam_data_prep/ops/subsetting.py
```diff
@@ -0,0 +1,57 @@
...
+                f"Could not find the all variables `{variables}` in the dataset. "
+                f"The available variables are {list(ds_input.data_vars)}"
+            ) from ex
+    else:
+        raise ValueError("The `variables` argument should be a list or a dictionary")
+
+    chunks = {
+        dim: chunking.get(dim, int(ds_subset[dim].count())) for dim in ds_subset.dims
+    }
+    ds_subset = ds_subset.chunk(chunks)
```
leifdenby, line 55: if the chunking is done here already (during load), why do we need to do it again later when deriving variables?

tests/test_derived_variables.py
leifdenby, line 1: I don't quite understand the mocking, patching and tests in here :) but let's talk about it so I can learn!

leifdenby (commented):
Really looking great! I will find a time in our calendar tomorrow so we can have a look together :)

mllam_data_prep/derived_variables.py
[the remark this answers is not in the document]
leifdenby: Yes, it was meant to be shorthand for "operations", sorry... regarding splitting this file it isn't so much object-oriented programming, but more about just splitting functions into separate modules that have different purposes. But let's talk about it and I'll show you what I mean :)

mllam_data_prep/create_dataset.py
```diff
@@ -120,23 +125,50 @@ def create_dataset(config: Config):
 
     output_config = config.output
     output_coord_ranges = output_config.coord_ranges
+    chunking_config = config.output.chunking
 
     dataarrays_by_target = defaultdict(list)
 
     for dataset_name, input_config in config.inputs.items():
         path = input_config.path
         variables = input_config.variables
```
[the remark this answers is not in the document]
ealerskans: Yes, I think that is a very good idea :)

mllam_data_prep/config.py
```diff
@@ -301,6 +328,54 @@ class Config(dataclass_wizard.JSONWizard, dataclass_wizard.YAMLWizard):
     class _(JSONWizard.Meta):
         raise_on_unknown_json_key = True
 
+    @staticmethod
+    def load_config(*args, **kwargs):
```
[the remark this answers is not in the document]
ealerskans: Really nice method! I have tried to implement it as suggested.

mllam_data_prep/create_dataset.py
ealerskans: The loop was inside the function (previously named `ops/loading/load_and_subset_dataset`) but then we were loading the dataset in there as well. I have now moved the looping outside the function (now called `ops/subsetting/extract_variable`) so that we are now checking the instance of `selected_varriables` and then extracting the selected variables. Please have a look and see what you think :)

mllam_data_prep/create_dataset.py
ealerskans: We now only return one variable at a time when calling `derive_variable` :)

mllam_data_prep/ops/derived_variables.py
ealerskans: I have updated the comment:

```
    # Chunking is needed for coordinates used to derive a variable since they are
    # not lazily loaded, as otherwise one might run into memory issues if using a
    # large dataset as input.
    # Any coordinates needed for the derivation, for which chunking should be performed,
    # should be converted to variables since it is not possible for *indexed* coordinates
    # to be chunked dask arrays
```

[See lines 62-67](https://github.com/mllam/mllam-data-prep/pull/34/files#diff-995872df159dd47379f25051fa53f065f92af26661e07bcddfa68ebd7a1d56e7R62-R67), so the explaining comment is a few lines above

mllam_data_prep/ops/derived_variables.py
ealerskans: I have tried to implement this suggestion! :)

mllam_data_prep/ops/derived_variables.py
ealerskans: I have updated the config file so that we now explicitly define which component of the time encodings ("sin" or "cos") we want to include. Also, I have added an extra section called `extra_kwargs` (suggestions for a better name?) in which one can specify function arguments that should not be extracted from the input dataset (such as `component` to specify if the sin or cos component should be extracted). I figured it could be nice to have since then the functions can take other arguments and not only variables from the dataset. Let me know if you know of a smarter way to do this :)

The `calculate_hour_of_day` and `calculate_day_of_year` functions now take an extra argument called `component` which should be either "sin" or "cos". This means that we now return only one variable - either the sin or cos component.

mllam_data_prep/ops/derived_variables.py
ealerskans: I have done something like this. See the answer to the comment above.

mllam_data_prep/ops/subsetting.py
ealerskans: Chunking is not done during loading anymore. I removed that since we're doing it on the whole dataset when saving it. But even if it were we would still have to do it in `derive_variable` since this chunking
```
ds_subset = ds_subset.chunk(chunks)
```

does not chunk the indexed coordinates, which in our case `time` is. And since `time` is used as input to `calculate_toa_radiation` we need to chunk it to not encounter a memory issue when running on the full dataset (because we are using `time`, `lat`, and `lon` to create a 3d field and since these are coordinates the computation is not lazy).

And just to make sure, I tested running `example.danra.yaml` using the v0.4.0 danra dataset as well and I didn't get any memory issues with the current code.

mllam_data_prep/ops/derived_variables.py
ealerskans: Thanks, I agree that it makes it clearer wherever possible :)

I have renamed `_check_for_required_attributes` to `_check_and_get_required_attributes` to make it clear that we here not only check that the required attributes are set but we also here handle how to do it when the attributes are defined both in the function and in the config file. Now it is also returning an `attrs` dictionary and we update the attributes in the calling function.

With respect to the other two functions:
```
def _return_dropped_coordinates(field, ds, required_coordinates, chunks):
    """
    ...
    """
    for req_coord in required_coordinates:
        if req_coord in chunks:
            field.coords[req_coord] = ds[req_coord]

    return field
```
adds the dropped coordinates back (coordinates such as `time` are dropped since otherwise they can't be chunked). I guess to make it clearer I could just skip the function and add the code instead of calling a function.

The other function
```
def _align_derived_variable(field, ds, target_dims):
    """
   ...
    """
    # Ensure that dimensions are ordered correctly
    field = field.transpose(
        *[dim for dim in target_dims if dim in field.dims], missing_dims="ignore"
    )

    # Add missing dimensions explicitly
    for dim in target_dims:
        if dim not in field.dims:
            field = field.expand_dims({dim: ds.sizes[dim]})

    # Broadcast to match only the target dimensions
    broadcast_shape = {dim: ds[dim] for dim in target_dims if dim in ds.dims}
    field = field.broadcast_like(xr.Dataset(coords=broadcast_shape))

    return field
```

aligns the dimensions of the derived variable with the dimensions of the dataset. Not sure how I would do this if the input argument `field` shouldn't be modified.

joeloskarsson:
Hi @ealerskans, just sneaking in a question about this. Me and @sadamov are eager to add on derived forcing inputs to some experiments. I see that there is still discussion ongoing above about the code structure and tests, but is all the actual functionality in place here? I.e. could we check out this branch and use it in the meantime to create the forcings we need?

ealerskans:
> Hi @ealerskans, just sneaking in a question about this. Me and @sadamov are eager to add on derived forcing inputs to some experiments. I see that there is still discussion ongoing above about the code structure and tests, but is all the actual functionality in place here? I.e. could we check out this branch and use it in the meantime to create the forcings we need?

Hi @joeloskarsson and @sadamov If you are only interested in adding the datetime forcings (hour of day and day of year) and top-of-the-atmosphere radiation), then you should be able to use this branch as it is now. I believe that I at one point also tested training neural-lam on a small test dataset which included derived variables as well. However no guarantees ;)

mafdmi:
> I have tried to update the PR now based on all the great suggestions and feedback I have gotten. What I think is missing now are
> 
>     * functionality of how to wrap a derived variable as an `xr.DataArray` if it is e.g. a `np.ndarray`
> 
>     * add more tests
> 
> 
> With respect to tests, I think that the following would be nice to test, but I am not sure how to do it:
> 
>     * That `_check_and_get_required_attributes` in `mllam_data_prep.ops.derive_variable.main` works as it should (i.e. how we handle attributes)
> 
>     * That `_get_derived_variable_function` in `mllam_data_prep.ops.derive_variable.main` works as it should (i.e. that we get the correct function, and that it also works for functions outside of `mllam_data_prep`)
> 
> 
> @mafdmi do you perhaps have any good ideas on 1) if it would be good to test them, and 2) how to do it?
> 
> Is the functionality to wrap in `xr.DataArray` something that should be included in this PR?
> 
> Then there is also the task of adding other derived fields, such as creating a land-sea mask or a boundary mask (https://github.com/mllam/neural-lam/blob/feature_calculate_forcings/create_forcings.py). But I guess these can be added in another PR.

Yes, I think we can test them. For `_check_and_get_required_attributes` I would test, that it returns the correct attributes depending on the different cases, e.g. if attribute is in field.attrs or not, the attrs is correctly updated. For `_get_derived_variable_function` we can setup tests to check that it can import various functions from the standard library, and test what happens with non-existing module. Do you want me to write a proposal for the tests or do want to give it a try from this?

ealerskans:
> > I have tried to update the PR now based on all the great suggestions and feedback I have gotten. What I think is missing now are
> > ```
> > * functionality of how to wrap a derived variable as an `xr.DataArray` if it is e.g. a `np.ndarray`
> > 
> > * add more tests
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With respect to tests, I think that the following would be nice to test, but I am not sure how to do it:
> > ```
> > * That `_check_and_get_required_attributes` in `mllam_data_prep.ops.derive_variable.main` works as it should (i.e. how we handle attributes)
> > 
> > * That `_get_derived_variable_function` in `mllam_data_prep.ops.derive_variable.main` works as it should (i.e. that we get the correct function, and that it also works for functions outside of `mllam_data_prep`)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @mafdmi do you perhaps have any good ideas on 1) if it would be good to test them, and 2) how to do it?
> > Is the functionality to wrap in `xr.DataArray` something that should be included in this PR?
> > Then there is also the task of adding other derived fields, such as creating a land-sea mask or a boundary mask (https://github.com/mllam/neural-lam/blob/feature_calculate_forcings/create_forcings.py). But I guess these can be added in another PR.
> 
> Yes, I think we can test them. For `_check_and_get_required_attributes` I would test, that it returns the correct attributes depending on the different cases, e.g. if attribute is in field.attrs or not, the attrs is correctly updated. For `_get_derived_variable_function` we can setup tests to check that it can import various functions from the standard library, and test what happens with non-existing module. Do you want me to write a proposal for the tests or do want to give it a try from this?

If you have time it would be really helpful if you could give it a go. Then I can try and learn from that :)

mafdmi:
> > Yes, I think we can test them. For `_check_and_get_required_attributes` I would test, that it returns the correct attributes depending on the different cases, e.g. if attribute is in field.attrs or not, the attrs is correctly updated. For `_get_derived_variable_function` we can setup tests to check that it can import various functions from the standard library, and test what happens with non-existing module. Do you want me to write a proposal for the tests or do want to give it a try from this?
> 
> If you have time it would be really helpful if you could give it a go. Then I can try and learn from that :)

I've opened https://github.com/ealerskans/mllam-data-prep/pull/1 with tests added:)

ealerskans:
@leifdenby I have implemented what we talked about in https://github.com/mllam/mllam-data-prep/pull/34/commits/0ecfccaecb04d6918cef66efe2a1b5ef199dd439.

mllam_data_prep/create_dataset.py
mafdmi, line 1: Not for this PR, but just to keep in the back of our heads. I think the `create_dataset` probably need some refactoring soon, since it is getting quite long:)

mllam_data_prep/create_dataset.py
```diff
@@ -126,23 +134,58 @@ def create_dataset(config: Config):
...
+                for var_name in selected_variables:
+                    ds[var_name] = extract_variable(ds=ds_input, var_name=var_name)
+            else:
+                raise ValueError(
+                    "The `variables` argument should be a list or a dictionary"
+                )
+
+        if derived_variables:
+            logger.info(f"Deriving variables from {dataset_name}")
+            for var_name, derived_variable in derived_variables.items():
```
mafdmi, line 182: Really nice!

mllam_data_prep/ops/derive_variable/main.py
```diff
@@ -0,0 +1,287 @@
...
+from loguru import logger
+
+from ..chunking import chunk_dataset
+
+REQUIRED_FIELD_ATTRIBUTES = ["units", "long_name"]
+
+
+def derive_variable(ds, derived_variable, chunking):
+    """
+    Load the dataset, and derive the specified variables
```
mafdmi, line 23: I think the "Load the dataset" is obsolete now

mllam_data_prep/ops/derive_variable/main.py
```diff
@@ -0,0 +1,287 @@
...
+            )
+            attrs[attribute] = expected_attributes[attribute]
+        else:
+            # Attributes are set in the function and nothing has been defined in the config file
+            attrs[attribute] = field.attrs[attribute]
+
+    return attrs
+
+
+def _return_dropped_coordinates(field, ds, required_coordinates, chunks):
```
mafdmi, line 222: Minor thing that is not necessarily something we need to change. When I read the name of this function again now, I got the impression, that it would return the dropped coordinates. But as I understand the function, it actually inserts the dropped coordinates into the already existing field dataset. So I just thought the function name might be misleading, but maybe I'm wrong.

mllam_data_prep/ops/subsetting.py
```diff
@@ -0,0 +1,50 @@
+def extract_variable(ds, var_name, coords_to_sample=dict()):
+    """
+    Extract specified variable from the provided the input dataset. If
```
mafdmi, line 3: Small typos: one "the" too much, and defines -> defined

README.md
```diff
@@ -315,11 +350,49 @@ The `inputs` section defines the source datasets to extract data from. Each sour
...
+- `derived_variables`: defines the variables to be derived from the variables available in the source dataset. This should be a dictionary where each key is the name of the variable to be derived and the value defines a dictionary with the following additional information. See also the 'Derived Variables' section for more details.
+  - `function`: the function used to derive a variable. This should be a string with the full namespace of the function, e.g. `mllam_data_prep.ops.derived_variables.physical_field.calculate_toa_radiation`.
+  - `kwargs`: arguments to `function`. This is a dictionary where each key is the named argument to `function` and each value is the input to the function. Here we distinguish between values to be extracted/selected from the input dataset and values supplied by the users themselves. Arguments with values to be extracted from the input dataset need to be prefixed with "input_dataset." to distinguish them from other arguments. See the 'Derived Variables' section for more details.
```
mafdmi, line 360: Typo: input_dataset -> ds_input

mafdmi (commented):
Looks great!! Only minor suggestions/fixes

mllam_data_prep/ops/derive_variable/main.py
ealerskans: Ah, perhaps the name is a bit misleading. I meant that the coordinates that were dropped from the field should be returned to the field.

This is also something Leif commented about before, that I should try to not modify `field` in the function since it is an argument to the function. However, I couldn't figure out a nice way to return it before but I think I can do that now. Then it would return the coordinates as you think. Then I could add it in the calling function instead. Should I give it a go? :) 

The same goes for `_align_derived_variable` but there I think it is more complicated not to modify `field` in the function even though it is an argument.

mafdmi: Ah, okay. Yes, maybe try to do that. It looks great for the other function where you did it:)

README.md
ealerskans: Solved in  97ee6dd.

mllam_data_prep/ops/subsetting.py
ealerskans: Solved in  97ee6dd.

mllam_data_prep/ops/derive_variable/main.py
ealerskans: Solved in  97ee6dd.

mllam_data_prep/ops/derive_variable/main.py
ealerskans: I decided to just skip the function since it only contained 3 lines and instead of calling `_return_dropped_coordinates` I added those lines directly in `derive_variable`. So I also removed the function since it was no longer needed.

I have not done anything to `_align_derived_variable`. This is still modifying the `field` argument and returning it.

sadamov:
> Hi @joeloskarsson and @sadamov If you are only interested in adding the datetime forcings (hour of day and day of year) and top-of-the-atmosphere radiation), then you should be able to use this branch as it is now. I believe that I at one point also tested training neural-lam on a small test dataset which included derived variables as well. However no guarantees ;)

This is brilliant, worked like a charm for my Swiss dataset! Thank you :pray:

mafdmi (approved):

mllam_data_prep/ops/derive_variable/main.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # Broadcast to match only the target dimensions
+    broadcast_shape = {dim: ds[dim] for dim in target_dims if dim in ds.dims}
+    field = field.broadcast_like(xr.Dataset(coords=broadcast_shape))
+
+    return field
+
+
+def get_latlon_coords_for_input(ds):
+    """Dummy function for getting lat and lon."""
```
leifdenby, line 258: Maybe we should instead write:

```suggestion
    """Placeholder function for getting latitude and longitude values. This will eventually be 
    replaced by routines handling proper projection support
    (see https://github.com/mllam/mllam-data-prep/pull/38)
    """
```

tests/derive_variable/test_main.py
leifdenby, line 1: really impressed with these tests!

leifdenby (approved):
I just have one suggestion for the docstring for that lat/lon function. Otherwise this is ready to merge 🥳

leifdenby (approved):

PR #35, opened 2024-11-07
Remove f-string from 'name_format' in config examples

ealerskans:
The new datastores functionality in neural-lam https://github.com/mllam/neural-lam/pull/66 assumes that the `name_format` of the features in the config are without the "f"-string.

With this addition, the example config and `README.md` in mllam-data-prep removes "f" from the `name_format` key to be consistent with that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ inputs:
       state_feature:
         method: stack_variables_by_var_name
         dims: [altitude]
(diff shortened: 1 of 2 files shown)


leifdenby (approved):
You are a 🌟 @ealerskans for doing this! Looks perfect to me.

Let's wait till the dev meeting next week so we can get this on the roadmap before we merge

PR #36, opened 2024-11-07
Remove global config for dataclass_wizard

leifdenby:
In working on the yaml configuration files in neural-lam I realised that setting global configuration for `dataclass_wizard` is a very bad idea. This should be obvious because different uses of `dataclass_wizard` might have different needs for configuration, however previously I didn't realise how to set the config when using `dataclass_wizard.YAMLWizard`. It turns out the key is to 1) also inherit from `dataclass_wizard.JSONWizard` (and make this the primary parent class) and 2) use `dataclass_wizard.JSONWizard.Meta` to define the config (not `dataclass_wizard.YAMLWizard.Meta` since this class does not exist).

This commit removes the global config for `dataclass_wizard` and only sets the configuration on `mllam-data-preps` `mllam_data_prep.config.Config` config dataclass.

### DIFF
--- a/mllam_data_prep/config.py
+++ b/mllam_data_prep/config.py
@@ -9,18 +9,6 @@ class InvalidConfigException(Exception):
     pass
 
 
-class GlobalJSONMeta(JSONWizard.Meta):
-    """
-    Global settings for the JSON load/dump process, that should apply to
-    *all* subclasses of `JSONWizard`.
-
-    Note: it does not matter where this class is defined, as long as it's
-    declared before any methods in `JSONWizard` are called.
-    """
-
-    raise_on_unknown_json_key = True
-
-
 @dataclass
 class Range:
     """
@@ -275,7 +263,7 @@ class Output:
 
 
 @dataclass
-class Config(dataclass_wizard.YAMLWizard):
+class Config(dataclass_wizard.JSONWizard, dataclass_wizard.YAMLWizard):
     """Configuration for the model.
 
     Attributes:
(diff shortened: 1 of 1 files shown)


leifdenby:
Hopefully the description above is clear @observingClouds, if not let's have a chat tomorrow. I would like for this to get into the next release, so I will bring it to the developer meeting next week.

observingClouds:
So if I understand correctly, the code here is about ensuring that all unkown parameters in the config files will raise an exception. Recently the `GlobalJSONMeta` class has been used with the major downside that also other modules loaded at runtime will be affected by this global setting if they use `dataclass_wizard`.

With this PR you ensure that only the local dataclasses are affected as it now becomes part of the `Config` dataclass and is no longer global.

Not being familiar with the `dataclass_wizard` the suggested change is clearer now.

observingClouds (approved):

leifdenby:
> So if I understand correctly, the code here is about ensuring that all unkown parameters in the config files will raise an exception. Recently the `GlobalJSONMeta` class has been used with the major downside that also other modules loaded at runtime will be affected by this global setting if they use `dataclass_wizard`.

That's exactly right, yes! Thanks again for reviewing this.

leifdenby:
> Merging this as per roadmap

Thanks! Although I hadn't written anything in the changelog yet 😄 But I will do that when I prepare the `v0.5.0` release

joeloskarsson:
Right, sorry about that :sweat_smile: Yes, please fix it with the 0.5.0 release. Maybe a good reason to have the PR template also here for MDP?

Issue #31, opened 2024-10-30
Introduce common input section that will be applied to all inputs


leifdenby:
# The problem

In working on the datastores functionality in neural-lam I needed the ability to define the projection information for the grid-coordinates in a zarr datasets created by `mllam-data-prep`. To do this a had a bit of a conundrum: where in the config to store this info? In the situation where all input datasets are given on the same projection it would be ideal to define this information only once in the config, but to make it clear that this information applies to every single source. Currently, it isn't possible to set config parameters "globally" for all inputs.

The proposition here is to add this _common_ section

For context I have currently made (on a separate branch https://github.com/mllam/mllam-data-prep/pull/18, which I would like to replace with a better implementation) what I feel like is a bit of a hack: adding an `extra` section to the config which the user can use to set things in the config that mllam-data-prep should ignore but can be used by other tools that read the same config. I don't really like this because this projection information is not used or even validated by `mllam-data-prep`. To add the domain-cropping that @TomasLandelius is planning to implement having this projection information understood by `mllam-data-prep` will also be key I think, so that `mllam-data-prep` knows how to get the latitude and longitude for each grid-point for each input dataset being used in a config.

# Proposition
Add a new optionally specially-named input called `__common__` the contents of which will be copied to all the inputs given.

So instead of writing:

```yaml
...
inputs:
  danra_height_levels:
    path: https://mllam-test-data.s3.eu-north-1.amazonaws.com/height_levels.zarr
    dims: [time, x, y, altitude]
    variables:
      u:
        altitude:
          values: [100,]
          units: m
      v:
        altitude:
          values: [100, ]
          units: m
    dim_mapping:
      time:
        method: rename
        dim: time
      state_feature:
        method: stack_variables_by_var_name
        dims: [altitude]
        name_format: f"{var_name}{altitude}m"
      grid_index:
        method: stack
        dims: [x, y]
    target_output_variable: state

  danra_surface:
    path: https://mllam-test-data.s3.eu-north-1.amazonaws.com/single_levels.zarr
    dims: [time, x, y]
    variables:
      # use surface incoming shortwave radiation as forcing
      - swavr0m
    dim_mapping:
      time:
        method: rename
        dim: time
      grid_index:
        method: stack
        dims: [x, y]
      forcing_feature:
        method: stack_variables_by_var_name
        name_format: f"{var_name}"
    target_output_variable: forcin
```

one could write
```yaml
...
inputs:
  __common__:
    dim_mapping:
[...]

PR #38, opened 2024-11-13
Projection support

### LINKED ISSUE #33: Make `mllam-data-prep` aware of geographical coordinates (by introducing projection info)

observingClouds:
eventually fixes #33

### DIFF
--- a/mllam_data_prep/create_dataset.py
+++ b/mllam_data_prep/create_dataset.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from pathlib import Path
 
 import numpy as np
+import pyproj
 import xarray as xr
 from loguru import logger
 from numcodecs import Blosc
@@ -12,6 +13,11 @@ from . import __version__
 from .config import Config, InvalidConfigException
 from .ops.loading import load_and_subset_dataset
 from .ops.mapping import map_dims_and_variables
+from .ops.projection import (
+    ProjectionInconsistencyWarning,
+    get_projection_crs,
+    validate_projection_consistency,
+)
 from .ops.selection import select_by_kwargs
 from .ops.statistics import calc_stats
 
@@ -106,6 +112,7 @@ def create_dataset(config: Config):
     output_coord_ranges = output_config.coord_ranges
 
     dataarrays_by_target = defaultdict(list)
+    projections = []
 
     for dataset_name, input_config in config.inputs.items():
         path = input_config.path
@@ -158,6 +165,10 @@ def create_dataset(config: Config):
 
         da_target.attrs["source_dataset"] = dataset_name
 
+        # get the projection information from the dataset and update it with the projection
+        # information given in the input config
+        projections.append(get_projection_crs(ds))
+
         # only need to do selection for the coordinates that the input dataset actually has
         if output_coord_ranges is not None:
             selection_kwargs = {}
@@ -167,6 +178,17 @@ def create_dataset(config: Config):
             da_target = select_by_kwargs(da_target, **selection_kwargs)
 
         dataarrays_by_target[target_output_var].append(da_target)
+
+    # validate projections across input datasets
+    try:
+        validate_projection_consistency(projections)
+    except ProjectionInconsistencyWarning as e:
+        logger.warning(f"Projection information might be ambiguous: {e}")
+    projection = pyproj.CRS.from_cf(set(projections).pop())
+
+    # TODO: generalize the retrieval of x and y coords
+    # coords = (dataarrays_by_target[target_output_var]['x'], dataarrays_by_target[target_output_var]['y'])
+    # lat, lon = get_latitude_longitude_from_projection(projection, coords)
 
     ds = _merge_dataarrays_by_target(dataarrays_by_target=dataarrays_by_target)
 
@@ -213,6 +235,11 @@ def create_dataset(config: Config):
             coords={"split_name": list(splits.keys()), "split_part": ["start", "end"]},
         )
         ds["splits"] = da_splits
+
+    # add the projection information to the dataset
+    ds["crs"] = projection.to_cf()
+    for var in ds.data_vars:
+        ds[var].attrs["grid_mapping"] = "crs"
 
     ds.attrs = {}
     ds.attrs["schema_version"] = config.schema_version
--- a/mllam_data_prep/ops/projection.py
+++ b/mllam_data_prep/ops/projection.py
@@ -156,7 +156,7 @@ def validate_projection_consistency(
     return
 
 
-def get_projection(ds: xr.Dataset) -> Dict[str, Any]:
+def get_projection_crs(ds: xr.Dataset) -> Dict[str, Any]:
     """Get the projection information from the dataset.
 
     Parameters
@@ -190,11 +190,11 @@ def get_projection(ds: xr.Dataset) -> Dict[str, Any]:
             f"Variables {vars_wo_proj} do not have a projection defined in the dataset {ds.encoding.get('source',None)}"
         )
 
-    proj_objs = {}
-    for proj in proj_vars:
-        proj_objs[proj] = pyproj.CRS.from_cf(ds[proj].attrs)
+    crss = {}
+    for crs in crss:
+        crss[crs] = ds[crs].attrs  # pyproj.CRS.from_cf(ds[proj].attrs)
 
-    return proj_objs
+    return crss
 
 
 def get_latitude_longitude_from_projection(
--- a/mllam_data_prep/ops/projection.py
+++ b/mllam_data_prep/ops/projection.py
@@ -177,7 +177,7 @@ def get_projection_crs(ds: xr.Dataset) -> Dict[str, Any]:
     """
     vars_w_proj = _get_projection_mappings(ds)
     proj_vars = set([proj for proj in vars_w_proj.values()])
-    vars_wo_proj = set(ds.variables) - set(vars_w_proj.keys()) - proj_vars
+    vars_wo_proj = set(ds.data_vars) - set(vars_w_proj.keys()) - proj_vars
 
     if len(proj_vars) > 1:
         raise ValueError(
--- a/mllam_data_prep/ops/projection.py
+++ b/mllam_data_prep/ops/projection.py
@@ -1,0 +1,231 @@
+"""Collection of operators handling projection information."""
+
+from typing import Any, Dict, List, Tuple, Union
+
+import pyproj
+import xarray as xr
+from pyproj import CRS
+
+
+class ProjectionInconsistencyWarning(Warning):
+    pass
+
+
+def _extract_grid_mapping_names(grid_mapping_attr: str) -> List[str]:
+    """Extract grid mapping names from the grid_mapping attribute.
+
+    Parameters
+    ----------
+    grid_mapping_attr : str
+        The grid_mapping attribute.
+
+    Returns
+    -------
+    list
+        List of grid mapping names.
+
+    Examples
+    --------
+    >>> m1 = "crsOSGB: x y crsWGS84: lat lon"
+    >>> _extract_grid_mapping_names(m1)
+    ['crsOSGB', 'crsWGS84']
+    >>> m2 = "crsOSGB"
+    >>> _extract_grid_mapping_names(m2)
+    ['crsOSGB']
+
+    Note: merge of https://github.com/pydata/xarray/pull/9765 of will allow xarray to handle this directly.
+    """
+    if ":" not in grid_mapping_attr:
+        return grid_mapping_attr.split(" ")
+    else:
+        return [
+            key.split(":")[0].strip()
+            for key in grid_mapping_attr.split(" ")
+            if ":" in key
+        ]
+
+
+def _get_projection_mappings(dataset: xr.Dataset) -> Dict[str, Any]:
+    """Get projections referenced by variables.
+
+    This function extracts the projection variables from a dataset
+    by evaluating the grid_mapping attribute of each variable following
+    the CF-Conventions.
+
+    Parameters
+    ----------
+    dataset : xr.Dataset
+        Dataset to get the projection information from
+
+    Returns
+    -------
+    Dict[str, str]
+        Dictionary with the variable names as keys and the projection
+        variable names as values
+    """
+    projections = {}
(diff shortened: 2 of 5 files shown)


mllam_data_prep/create_dataset.py
```diff
@@ -168,6 +198,18 @@ def create_dataset(config: Config):
...
+    if projections:
+        try:
+            validate_projection_consistency(projections)
+        except ProjectionInconsistencyWarning as e:
+            logger.warning(f"Projection information might be ambiguous: {e}")
+        projection = pyproj.CRS.from_cf(projections[0])
+
+        # TODO: generalize the retrieval of x and y coords
+        # coords = (dataarrays_by_target[target_output_var][0]['x'], dataarrays_by_target[target_output_var][0]['y'])
+        # lon, lat = get_latitude_longitude_from_projection(projection, coords)
```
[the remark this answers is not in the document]
observingClouds: the function is available from `from .ops.projection`

mllam_data_prep/create_dataset.py
```diff
@@ -127,6 +135,10 @@ def create_dataset(config: Config):
             dataset_name=dataset_name,
         )
 
+        # TODO: remove this once grid_mapping is set in config file
+        for var in ds.data_vars:
+            ds[var].attrs["grid_mapping"] = "crs"
+
```
observingClouds, line 141: @leifdenby this sets the `grid_mapping` attribute for now here and so we can postpone the discussion on the structure of the `variables` section in the config file

mllam_data_prep/ops/projection.py
```diff
@@ -0,0 +1,241 @@
...
+    Examples
+    --------
+    >>> m1 = "crsOSGB: x y crsWGS84: lat lon"
+    >>> _extract_grid_mapping_names(m1)
+    ['crsOSGB', 'crsWGS84']
+    >>> m2 = "crsOSGB"
+    >>> _extract_grid_mapping_names(m2)
+    ['crsOSGB']
+
+    Note: merge of https://github.com/pydata/xarray/pull/9765 of will allow xarray to handle this directly.
```
SimonKamuk, line 36: nitpicking typo ;)
```suggestion
    Note: merge of https://github.com/pydata/xarray/pull/9765 will allow xarray to handle this directly.
```

SimonKamuk (commented):
Great work! Only some small comments from me :)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- add support for reading, setting and outputting CF-Conventions complient projection information. ![\#38](https://github.com/mllam/mllam-data-prep/pull/38)
```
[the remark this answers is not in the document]
observingClouds: I adjusted the CHANGELOG.md. Thanks!

mllam_data_prep/create_dataset.py
```diff
@@ -158,6 +170,24 @@ def create_dataset(config: Config):
...
+            logger.warning(
+                f"Projection information not found in dataset {dataset_name}, using projection information from config."
+            )
+            projection_crs = {
+                crs_var: p.attributes for crs_var, p in dataset_projections.items()
+            }
+        elif projection_crs is None and not dataset_projections:
+            logger.warning(
+                f"No projection information found neither in dataset {dataset_name}, nor in config."
+            )
```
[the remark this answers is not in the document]
observingClouds: Done

leifdenby:
@observingClouds I just had a quick read of this again and I think maybe we should sit down and talk it through. There are somethings you have left "todo", which might be because you're missing something from me. So I will try and catch you before the holidays :)

mllam_data_prep/create_dataset.py
```diff
@@ -184,6 +218,18 @@ def create_dataset(config: Config):
...
+    if projections:
+        try:
+            validate_projection_consistency(projections)
+        except ProjectionInconsistencyWarning as e:
+            logger.warning(f"Projection information might be ambiguous: {e}")
+        projection = pyproj.CRS.from_cf(projections[0])
+
+        # TODO: generalize the retrieval of x and y coords
+        # coords = (dataarrays_by_target[target_output_var][0]['x'], dataarrays_by_target[target_output_var][0]['y'])
+        # lon, lat = get_latitude_longitude_from_projection(projection, coords)
```
observingClouds, line 231: @leifdenby this todo was just an example on where the projection information is available to derive lat and lon. I can delete this for now as this is not yet supported with mllam-data-prep

mllam_data_prep/create_dataset.py
```diff
@@ -143,6 +151,10 @@ def create_dataset(config: Config):
             dataset_name=dataset_name,
         )
 
+        # TODO: remove this once grid_mapping is set in config file
+        for var in ds.data_vars:
+            ds[var].attrs["grid_mapping"] = "__common__"
+
```
observingClouds, line 157: @leifdenby this is just a remark for the future when we allow the setting of variable specific projections or read the projection information directly from the input file

observingClouds: We can delete this and mention it maybe in an issue so we do not forget.

leifdenby:
Ok, thanks for the chat @observingClouds!

Here's what we decided (please let me know if I missed something / wrote it incorrectly):

- [ ] check that the parsed projection (i.e. the produced `pyproj.CRS` object) has the bbox attribute defined. The bounding box is required as if the resulting projection object is to be used by cartopy for plotting (otherwise cartopy cannot use it). Issue a warning if the user hasn't included the bbox (we can explicitly raise an exception in neural-lam if this is missing and tell the user to add it).
- [ ] include only `crs_wkt` string to define projection in config. This means that all other projection definition components should be removed.  In the CF-conventions the `crs_wkt` string takes precedence anyway and is the only way to define a bbox (which is needed to create a projection for plotting with cartopy, see above). To simplify the implementation here we only use the `crs_wkt` string.
- [ ] Implement config format which a) does not use a `__common__` section (as suggested in #31), we decided it was better to be explicit about the projection definition rather than implicit. This means that the user must copy the same projection to all input datasets in the config, b) names the projection explicitly (here called `danra_projection`) which follows the CF-compliant approach where projections are explicitly named, c) does not require that the named projection to be used is explicitly named for each variable selected (we can add that in future, but for now we will implement that one and only one projection per input dataset may be given, and that is assigned to each variable):
```yaml
inputs:
  danra_height_levels:
    path: https://mllam-test-data.s3.eu-north-1.amazonaws.com/height_levels.zarr
    dims: [time, x, y, altitude]
    variables:
      u:
        altitude:
          values: [100,]
          units: m
      v:
        altitude:
          values: [100, ]
          units: m
    projections:
      danra_projection:
        crs_wkt: 'PROJCRS["DMI HARMONIE DANRA lambert projection", BASEGEOGCRS["DMI HARMONIE DANRA lambert CRS", DATUM["DMI HARMONIE DANRA lambert datum", ELLIPSOID["Sphere", 6367470, 0, LENGTHUNIT["metre", 1, ID["EPSG", 9001]]]], PRIMEM["Greenwich", 0, ANGLEUNIT["degree", 0.0174532925199433, ID["EPSG", 8901]]], ID["EPSG",4035]], CONVERSION["Lambert Conic Conformal (2SP)", METHOD["Lambert Conic Conformal (2SP)", ID["EPSG", 9802]], PARAMETER["Latitude of false origin", 56.7, ANGLEUNIT["degree", 0.0174532925199433, ID["EPSG", 8821]]], PARAMETER["Longitude of false origin", 25, ANGLEUNIT["degree", 0.0174532925199433, ID["EPSG", 8822]]], PARAMETER["Latitude of 1st standard parallel", 56.7, ANGLEUNIT["degree", 0.0174532925199433, ID["EPSG", 8823]]], PARAMETER["Latitude of 2nd standard parallel", 56.7, ANGLEUNIT["degree", 0.0174532925199433, ID["EPSG", 8824]]], PARAMETER["Easting at false origin", 0, LENGTHUNIT["metre", 1, ID["EPSG", 8826]]], PARAMETER["Northing at false origin", 0, LENGTHUNIT["metre", 1, ID["EPSG", 8827]]]], CS[Cartesian, 2], AXIS["(E)", east, ORDER[1], LENGTHUNIT["metre", 1, ID["EPSG", 9001]]], AXIS["(N)", north, ORDER[2], LENGTHUNIT["metre", 1, ID["EPSG", 9001]]], USAGE[AREA["Denmark and surrounding regions"], BBOX[47, -3, 65, 25], SCOPE["Danra reanalysis projection"]]]'
        dims: [x, y]

    dim_mapping:
      time:
        method: rename
        dim: time
      state_feature:
        method: stack_variables_by_var_name
        dims: [altitude]
        name_format: f"{var_name}{altitude}m"
    target_output_variable: state
```
- [ ] add setting of CF-compliant projection info on output of `mllam-data-prep` by defining the projection in a CF-compliant manner (again with `crs_wkt`) and setting the `grid_mapping` attr on each output variable (that contains the dimensions that the projection uses) to reference that grid-mapping. It will likely also be necessary to set the coordinate names for coordinates used in the projection in a CF-compliant manner too.
- [ ] Write short section in README on 1) how to add projection info using the new section in the config, 2) the importance of the bbox, 3) may encourage people to raise an issue if they are unsure how to make their own crs wkt string?

observingClouds:
@leifdenby now that we make `projections` a required argument of the config file, I wonder how we make this best work with some of the tests against older schema, e.g.:

```
>   ???
E   dataclass_wizard.errors.MissingFields: `InputDataset.__init__()` missing required fields.
E     Provided: ['path', 'dims', 'variables', 'dim_mapping', 'target_output_variable']
E     Missing: ['projections']
E     Input JSON: {"path": "https://mllam-test-data.s3.eu-north-1.amazonaws.com/height_levels.zarr", "dims": ["time", "x", "y", "altitude"], "variables": {"u": {"altitude": {"values": [100], "units": "m"}}, "v": {"altitude": {"values": [100], "units": "m"}}}, "dim_mapping": {"time": {"method": "rename", "dim": "time"}, "state_feature": {"method": "stack_variables_by_var_name", "dims": ["altitude"], "name_format": "f\"{var_name}{altitude}m\""}, "grid_index": {"method": "stack", "dims": ["x", "y"]}}, "target_output_variable": "state"}
E     error: InputDataset.__init__() missing 1 required positional argument: 'projections'

<string>:36: MissingFields
============================================================ short test summary info ============================================================
FAILED tests/test_from_config.py::test_config_revision_examples[fp_example0] - dataclass_wizard.errors.MissingFields: `InputDataset.__init__()` missing required fields.
```

leifdenby:
This is still great work! But given the short timeline with Gefion right now and @joeloskarsson and @sadamov are working intensely to train models I think we should remove this from the roadmap for now (as we would otherwise be introducing a breaking change). I checked with @observingClouds and he's ok with that :)

PR #42, opened 2024-11-21
Feature/add more statistics

mafdmi:
Reworked the way we define statistical methods and the way we invoke them from the *.yaml file.

Added new statistical methods to calculate:
- mean and std across time only (no mean/std over grid_index)
- mean and std across diurnal cycles and grid_index
- mean and std across diurnal cycles only

Fixed typos in README. Removed duplicate and obsolete attributes section in docstrings.

The new way to specify compute statistics in the *.yaml file looks like the following:
```yaml
compute_statistics:
    MeanOperator:
      - var_name: mean
        dims: [grid_index, time]
    StdOperator:
      - var_name: std
        dims: [grid_index, time]
    DiffMeanOperator:
      - var_name: diff_mean
        dims: [grid_index, time]
      - var_name: diff_time_mean
        dims: [time]
    ...
```
and follows the structure
```yaml
compute_statistics:
    <name-of-python-class>:
      - var_name: <varible-name1>
        dims: [<dim1>, <dim2>, ...]
      - var_name: <varible-name2>
        dims: [<dim2>, <dim3>, ...]
```
where
- <name-of-python-class> shall be the name of a subclass of `ops.statistics::StatisticOperator` defined in `ops/statistics.py`
- <varible-name*> is a variable name to be used when saving the statistics to the dataset
- the list of dims can vary between <varible-name*> items.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+.vscode/
 
 # C extensions
 *.so
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ python -m pip install mllam-data-prep[dask-distributed]
 
 ## Developing `mllam-data-prep`
 
-To work on developing `mllam-data-prep` it easiest to install and manage the dependencies with [pdm](https://pdm.fming.dev/). To get started clone your fork of [the main repo](https://github.com/mllam/mllam-data-prep) locally:
+To work on developing `mllam-data-prep` it is easiest to install and manage the dependencies with [pdm](https://pdm.fming.dev/). To get started clone your fork of [the main repo](https://github.com/mllam/mllam-data-prep) locally:
 
 ```bash
 git clone https://github.com/<your-github-username>/mllam-data-prep
@@ -41,7 +41,7 @@ pdm use --venv in-project
 pdm install
 ```
 
-All the linting is handelled by `pre-commit` which can be setup to automatically be run on each `git commit` by installing the git commit hook:
+All the linting is handled by `pre-commit` which can be setup to automatically be run on each `git commit` by installing the git commit hook:
 
 ```bash
 pdm run pre-commit install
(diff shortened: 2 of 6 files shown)


mafdmi:
Maybe `var_name` is not the best, since it is actually more a string that is appended to the full variable name to which the statistics is saved to in the dataset. What do you think?

SimonKamuk:
Great functionality to add! I agree that var_name is not the most descriptive, maybe something more along the lines of statistic_name? Or it could be inferred from the supplied arguments? In any case, there still seems to be some tests failing.

I guess this will be a breaking change, since existing yaml files well need to be modified. What about the datasets outputted from mllam-data-prep, will they be changed? As long as var_name (or whatever you end up calling it) matches, then I guess the output stays the same?

observingClouds:
One might also checkout https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#statistics-more-than-one-axis and add the additional `cell-method` attribute so one knows over which dimensions the statistics have been taken.

observingClouds:
> var_name is not the best, since it is actually more a string that is appended to the full variable name

The question for me would be how we can read these statistics values in neural-lam. Appending the variable names by a string can be tricky to disentangle again. A delimiter like `.` might help. Then I would suggest instead of `var_name`, `varname_suffix` or `statistics_suffix`

If the output dimensions of the statistics are the same across datasets, I could also imagine, statistic variables that have the actual variable names as dimension:
```
float mean(time, varname):
    ....
float RMSE(time, varname):
    ....
```

mafdmi:
> One might also checkout https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#statistics-more-than-one-axis and add the additional `cell-method` attribute so one knows over which dimensions the statistics have been taken.

Good suggestion! Would be good to use standards for this.

observingClouds:
@mafdmi here are even more useful resources on how to create the variables names: https://cfconventions.org/Data/cf-standard-names/docs/guidelines.html#id2798144

leifdenby:
Closing since this work is continuing in a separate package on the following repo: https://github.com/mllam/mllam-verification

PR #45, opened 2024-12-09
Add support for cropping domain with padded lat/lon convex hull

leifdenby:
## Describe your changes

The aim of this PR is to add support for cropping one dataset using the convex hull of lat/lon coordinates of grid-points in another dataset. 

This is motivated by the need when doing LAM (limited area) based forecasting to be able to define forcing input in a padding region around the limited area. This could for example be when training on DANRA reanalysis one might want to force on the boundary with ERA5 reanalysis. This work builds on @TomasLandelius proof-of-concept implementation (https://github.com/mllam/weather-model-graphs/issues/30) which takes the convex hull of the lat/lon coordinates of grid-points in the limited area (e.g. DANRA) domain, splits the points in the boundary domain by whether they are interior or external to this convex hull (again using the lat/lon coordinates) and defines points in the boundary region as those that are within a given distance (in radians from Earth's center) to the closest point within the convex hull.

So far I have taken the implementation from @TomasLandelius and structured into a notebook (and used DANRA and ERA5 zarr datasets in cloud storage).

Below I have made an example where ERA5 sample points are selected that are within 0.2 radians of the convex hull of grid-points in DANRA:

output

For the lat/lon convex hull calculations the [spherical-geometry](https://github.com/spacetelescope/spherical_geometry) package is required. For plotting of course `matplotlib` and `cartopy`. I am thinking of putting the visualiation related packages in an optional `[visualisation]` extra requirements, but I am not sure if it makes sense to always require the spherical-geometry package. 

## Issue Link

Implements https://github.com/mllam/weather-model-graphs/issues/30

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)

- [ ] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [ ] I have performed a self-review of my code
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [ ] I have requested a reviewer and an assignee (assignee is responsible for merging)

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

- [ ] I have added a line to the CHANGELOG describing this change, in a section

## Checklist for assignee

- [ ] PR is up to date with the base branch
- [ ] the tests pass
- [ ] author has added an entry to the changelog (and designated the change as *added*, *changed* or *fixed*)


TomasLandelius:
Looks great!

/Tomas


-------- Originalmeddelande --------
Från: Leif Denby ***@***.***>
Datum: 2024-12-10 20:49 (GMT+01:00)
Till: mllam/mllam-data-prep ***@***.***>
Kopia: Landelius Tomas ***@***.***>, Mention ***@***.***>
Ämne: Re: [mllam/mllam-data-prep] Add support for cropping domain with padded lat/lon convex hull (PR #45)


This is getting there. Added some tests now so I can ensure the different steps of the algorithm works. Here's a nice plot for the convex-hull interior/exterior mask:

billede.png (view on web)<https://github.com/user-attachments/assets/087309b0-7ffb-4e13-afd7-1054d01f4330>

Separating out the mask calculation will also make it possible to add a cropping method that just crops using inside/outside the convex hull in future

leifdenby:
> This is looking great! What's your idea for how to specify the dataset to base the mask on (i.e. where to get the coordinates from the interior region from)? Would that be an already processed mdp zarr, or any arbitrary zarr? Should the path to that be specified in the mdp config file file the boundary dataset?

Thank you :)

What I have done so far is adapt the `output` part of the config to make it possible to point the path of a config for a dataset that should be treated as the *interior* domain. Like so:

```yaml
output:
  ...
  domain_cropping:
    margin_width_degrees: 0.2
    interior_dataset_config_path: example.danra.yaml
```
https://github.com/mllam/mllam-data-prep/pull/45/files#diff-33f36098ed98a22bbf317f860675cc821a3ccbb56e1650f087c8cc4b3b245038R30

I adapted @TomasLandelius code so that one can define the angle in degrees instead of radians (since I think that is a bit more intuitive to reason about). And I've left space for the possibility for other cropping methods to be defined later (for example one could use the convex hull to make two overlapping domains). I am very open to comments on the changes above though :)

Also, the test for DANRA + weatherbench ERA5 currently faisl because #33 isn't complete yet. I think it would be get that one in first. But other than that it is complete 🥳

docs/domain-cropping.ipynb
joeloskarsson, line 1: This file is great, however it is mainly a documentation of the implementation of domain-cropping, rather than a documentation of how to do it in practice (for a user). It is still good to have, but I think this might be good to clarify in here, that the user should not do something like what's in the file, but actually just specify a line in the config.

joeloskarsson, line 1: Document also the new config fields in the README

mllam_data_prep/ops/cropping.py
```diff
@@ -0,0 +1,201 @@
...
+        {grid_index_dim: "grid_index_ref"}
+    )
+
+    # create a mask from the convex hull of ds_reference for the grid points in ds
+    da_ch_mask = create_convex_hull_mask(
+        ds=ds, ds_reference=ds_reference_separate_gridindex
+    )
+
+    # only consider points that are external to the convex hull
+    ds_exterior = ds.where(~da_ch_mask, drop=True)
```
joeloskarsson, line 133: Now this always removes points that are within the convex hull. I would really want the functionality to still keep these (overlapping regions). It seems that that should not be a major step to implement as an option here? If not, I think we could try to do that in this PR as well. Just need to figure out a useful way to configure that in the yaml-file.

mllam_data_prep/ops/cropping.py
```diff
@@ -0,0 +1,201 @@
...
+    -------
+    tuple (xarray.DataArray, xarray.DataArray)
+        The latitude and longitude coordinates.
+    """
+    if "latitude" in da.coords and "longitude" in da.coords:
+        return (da.latitude, da.longitude)
+    elif "lat" in da.coords and "lon" in da.coords:
+        return (da.lat, da.lon)
+    else:
+        import ipdb
```
joeloskarsson, line 26: ipdb is now only a dev-dependency, but if it should be used here it needs to be a general dependency (or latlon-domain-crop)

example.era5_cropped.yaml
```diff
@@ -0,0 +1,109 @@
...
+          dims: [grid_index, time]
+      val:
+        start: 1990-09-06T00:00
+        end: 1990-09-07T00:00
+      test:
+        start: 1990-09-07T00:00
+        end: 1990-09-09T00:00
+  domain_cropping:
+    margin_width_degrees: 0.2
+    interior_dataset_config_path: example.danra.yaml
```
joeloskarsson, line 32: Would it not be possible to have this (optionally) point to an already saved zarr, instead of a config file? If I understand correctly, this now rebuilds the whole reference `xr.Dataset` from the config given here, and uses that for the cropping. I understand that this doesn't actually download or save all the data again, but it still seems a bit unneccesary to do all the Xarray operations again, given that this would almost always be used after the reference data has already been processed and saved to disk as zarr.

mllam_data_prep/ops/cropping.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Returns
+    -------
+    tuple (xarray.DataArray, xarray.DataArray)
+        The latitude and longitude coordinates.
+    """
+    if "latitude" in da.coords and "longitude" in da.coords:
+        return (da.latitude, da.longitude)
+    elif "lat" in da.coords and "lon" in da.coords:
+        return (da.lat, da.lon)
```
joeloskarsson, line 24: Better to return longitude, latitude ordering to be consistent with x, y ordering of rest of codebase.

mllam_data_prep/ops/cropping.py
```diff
@@ -0,0 +1,201 @@
...
+        The dataset for which to create the mask.
+    ds_reference : xarray.Dataset
+        The reference dataset from which to create the convex hull of the coordinates.
+
+    Returns
+    -------
+    xarray.DataArray
+        A boolean mask indicating which points are interior to the convex hull.
+    """
+    da_lat, da_lon = _get_latlon_coords(ds)
```
joeloskarsson, line 50: When testing this out I get an error here. Now the example ERA5 config renames latitude+longitude to y+x, which I think is needed to make the output work with neural-lam? It does however mean that there are no lat/lon in the coords of the dataset here.

mllam_data_prep/ops/cropping.py
```diff
@@ -0,0 +1,201 @@
...
+        ds=ds, ds_reference=ds_reference_separate_gridindex
+    )
+
+    # only consider points that are external to the convex hull
+    ds_exterior = ds.where(~da_ch_mask, drop=True)
+
+    da_xyz = _latlon_to_unit_sphere_xyz(ds_exterior.lat, ds_exterior.lon)
+    da_xyz_ref = _latlon_to_unit_sphere_xyz(
+        ds_reference_separate_gridindex.lat, ds_reference_separate_gridindex.lon
+    )
```
joeloskarsson, line 138: Here `.lat` and `.lon` is explicitly used, putting a requirement on these exact names being present in the dataset. This is strange as in `create_convex_hull_mask` the function `_get_latlon_coords` is used to not make this assumption. This also causes an error for the same reason as above, that these coordinates will already be renamed to x/y at this point.

mllam_data_prep/ops/cropping.py
joeloskarsson: Diff for local changes I had to do to get it running. With this I got rid of the errors and was able to save an ERA5 zarr.

```
--- a/mllam_data_prep/ops/cropping.py
+++ b/mllam_data_prep/ops/cropping.py
@@ -22,6 +22,8 @@ def _get_latlon_coords(da: xr.DataArray) -> tuple:
         return (da.latitude, da.longitude)
     elif "lat" in da.coords and "lon" in da.coords:
         return (da.lat, da.lon)
+    elif "y" in da.coords and "x" in da.coords:
+        return (da.y, da.x)
     else:
         import ipdb
 
@@ -132,9 +134,9 @@ def distance_to_convex_hull_boundary(
     # only consider points that are external to the convex hull
     ds_exterior = ds.where(~da_ch_mask, drop=True)
 
-    da_xyz = _latlon_to_unit_sphere_xyz(ds_exterior.lat, ds_exterior.lon)
+    da_xyz = _latlon_to_unit_sphere_xyz(ds_exterior.y, ds_exterior.x)
     da_xyz_ref = _latlon_to_unit_sphere_xyz(
-        ds_reference_separate_gridindex.lat, ds_reference_separate_gridindex.lon
+        ds_reference_separate_gridindex.y, ds_reference_separate_gridindex.x
     )
```

mllam_data_prep/ops/cropping.py
```diff
@@ -0,0 +1,201 @@
...
+        val = np.min(np.arccos(dotproduct))
+        return val
+
+    da_mindist_to_ref = xr.apply_ufunc(
+        calc_dist,
+        da_xyz,
+        input_core_dims=[["xyz"]],
+        output_core_dims=[[]],
+        vectorize=True,
+    )
```
joeloskarsson, line 151: I'm scared about the scaling of this, being $O(NM)$ (number of points in both sets). I understand that this can be parallelized, but we can do better algorithmically.

Once we have computed the convex hull, why don't we compute the distance to this polygon rather than to all points? This does not seem directly implemented in `spherical_geometry`, but maybe https://spherely.readthedocs.io/en/latest/_api_generated/spherely.distance.html#spherely.distance does this correctly?

Alternatively, if we can't measure the distance to the hull, can we do some kd-tree for this, e.g. with https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html + haversine distance?

mllam_data_prep/ops/cropping.py
joeloskarsson: Above makes the code run with the config, but the distance computation then breaks. I can't figure out why, but I think it relates to mixing of xy and latlon coordinates somewhere.

mllam_data_prep/create_dataset.py
```diff
@@ -230,6 +239,26 @@ def create_dataset(config: Config):
...
+        domain_cropping = config.output.domain_cropping
+        ds_interior_domain = create_dataset(config=config_interior_domain)
+        logger.info(
+            f"Cropping dataset to within convex hull margin of {ds_interior_domain} with a margin of {domain_cropping.margin_width_degrees} degrees"
+        )
+        ds = crop_to_within_convex_hull_margin(
+            ds=ds,
+            ds_reference=ds_interior_domain,
+            max_dist=domain_cropping.margin_width_degrees,
+        )
```
joeloskarsson, line 260: At the moment this seems to introduce a `grid_index` coordinate for the `ds["splits"]` variable, that should not be there. This also causes errors downstream in neural-lam.

joeloskarsson: Digging a bit more here I am not sure if it actually is the cropping that causes this. It still is an issue, but I can't figure out what causes it.

docs/domain-cropping.ipynb
leifdenby: Yes, good point. I've mostly left this in here because I was using it during development and I thought the examples might be useful later. Maybe calling it documentation isn't quite right. In general though it would be nice with some documentation on how to use `mllam_data_prep` not from the command line but by importing the package. I haven't had a chance to write that yet, but this notebook was supposed to be a start along those lines.

example.era5_cropped.yaml
leifdenby: Yes, in theory we could do that. I was a bit in doubt too about recomputing. But until we have a way to save in the produced dataset the config the dataset was built from (which is something we have previously discussed, but I don't think it has been worked on yet) then I wasn't too sure about doing this (this feels like cache invalidation, which is always a tricky thing to get right). How about I add functionality which checks for a zarr dataset named after the config (e.g. `example.danra.zarr` for `example.danra.yaml`) and if it exists I use that instead but issue a warning to the user that we are not checking the contents? I can then remove that warning once we have functionality in place to check that the dataset on disk actually is made from the config file that we were hoping for.

mllam_data_prep/ops/cropping.py
TomasLandelius: We can get the points defining the polygon from `chull_lam_lon, chull_lam_lat = list(chull_lam.to_lonlat())[0]`. Then we only need to compute minimum distance from (a reasonable subset of) the points in the border to the arcs defined by the polygon segments defining the hull. I think the minimum distance can be computed with something like:
```
def shortest_distance_to_arc(P_lat, P_lon, A_lat, A_lon, B_lat, B_lon, radius=6371):
    P = lat_lon_to_cartesian(P_lat, P_lon)
    A = lat_lon_to_cartesian(A_lat, A_lon)
    B = lat_lon_to_cartesian(B_lat, B_lon)
    
    # Calculate normal vector to the plane of the great circle
    n = np.cross(A, B)
    n = n / np.linalg.norm(n)  # Normalize
    
    # Project P onto the plane
    P_proj = P - np.dot(P, n)[:, np.newaxis] * n
    
    # Normalize to get Q on the sphere's surface
    Q = P_proj / np.linalg.norm(P_proj, axis=1)[:, np.newaxis]
    
    # Calculate the angle between P and Q
    theta = np.arccos(np.clip(np.sum(P * Q, axis=1), -1, 1))
    
    # Check if Q is between A and B
    Q_between = (np.dot(np.cross(A, Q), n) >= 0) & (np.dot(np.cross(Q, B), n) >= 0)
    
    # Calculate distances PA and PB
    PA = np.arccos(np.clip(np.dot(P, A), -1, 1))
    PB = np.arccos(np.clip(np.dot(P, B), -1, 1))
    
    # Choose the appropriate distance
    distances = np.where(Q_between, theta, np.minimum(PA, PB))
    
    return radius * distances
```
Some technical admin is needed to loop over the border point and the segments and find the minimum. Worth the effort?

mllam_data_prep/ops/cropping.py
joeloskarsson: I'm not sure what you mean above, I am just talking about the ordering of return arguments
```suggestion
        return (da.longitude, da.latitude)
    elif "lat" in da.coords and "lon" in da.coords:
        return (da.lon, da.lat)
```
(and swapping that ordering in later assignments that call this function)

mllam_data_prep/ops/cropping.py
leifdenby: Yes, agreed. It would be good to optimise this. I was going for creating a working implementation with what we have first, and then a later PR could do the extra work to optimise parts of this.

leifdenby: I was thinking that I would take quick look to see if I could get the just the coordinates of the polygon though (as you also suggested) and implement that here if it turns out to be easy

mllam_data_prep/create_dataset.py
leifdenby: I will look into it

mllam_data_prep/ops/cropping.py
leifdenby: yes, these direct referencing of `.lat` and `.lot` here was a mistake. I have removed this in commit https://github.com/mllam/mllam-data-prep/pull/45/commits/ba467183ec88959aa529fc0574daf1c82b5d1a1c

mllam_data_prep/ops/cropping.py
leifdenby: Thanks! Assuming that the latitude and longitude values are available as `.lat` and `.lot` on the dataset was a mistake (I meant to wrap all the access to the latitude and longitude coordinates with the `_get_latlon_coords` function, but missed this one. If I had implemented the testing before I would have caught this...)

The code in neural-lam actually sets the cartesian coordinates (or maybe they should be called "regular grid coordinates") when loading the dataset: https://github.com/mllam/neural-lam/blob/main/neural_lam/datastore/mdp.py#L108 and uses the config to work out the name of the coordinates that were stacked. So the coordinates that were stacked don't have to be renamed to `x` and `y` (that is the idea anyway)

mllam_data_prep/ops/cropping.py
joeloskarsson: I think that just getting the coordinates of the points that make up the polygon and restricting the distance computation to those would be an optimization that's completely sufficient. Great if you can give that a try. If it ends up being too complicated I am also fine with just optimizing this later.

mllam_data_prep/ops/cropping.py
leifdenby: Ah I see :) I thought you meant the order of the coordinates with the `xr.DataArray`s that are returned. Ok yes, we can change that, I just thought it made sense to have that order since people say "lat/lon-values", but we can make it explicit that the longitude are the first elements of the tuple

mllam_data_prep/ops/cropping.py
leifdenby: Ok, I've got cropping including the interior done now:

billede

billede

mllam_data_prep/ops/cropping.py
joeloskarsson: Extracting the points that make up the convex hull polygon in spherical_geometry and computing the distance to those points does not work. For the DANRA domain as example that polygon only has points at the corners (blue), giving a weird cropping of lat-lon ERA data (red)

chull_points

(It did speed up the cropping 10000 times though :laughing:). So we would need to also compute the distance to the actual edges of the polygon.

TomasLandelius:
You need to calculate the distance to the polygon, not just it's edges. It should be possible by considering all lines (great arcs) between the points making up the polygon.


-------- Originalmeddelande --------
Från: Joel Oskarsson ***@***.***>
Datum: 2025-01-09 18:08 (GMT+01:00)
Till: mllam/mllam-data-prep ***@***.***>
Kopia: Landelius Tomas ***@***.***>, Mention ***@***.***>
Ämne: Re: [mllam/mllam-data-prep] Add support for cropping domain with padded lat/lon convex hull (PR #45)


@joeloskarsson commented on this pull request.

________________________________

In mllam_data_prep/ops/cropping.py<https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1909178516>:

> +    def calc_dist(da_pt_xyz):
+        dotproduct = np.dot(da_xyz_ref, da_pt_xyz.T)
+        val = np.min(np.arccos(dotproduct))
+        return val
+
+    da_mindist_to_ref = xr.apply_ufunc(
+        calc_dist,
+        da_xyz,
+        input_core_dims=[["xyz"]],
+        output_core_dims=[[]],
+        vectorize=True,
+    )


Extracting the points that make up the convex hull polygon in spherical_geometry and computing the distance to those points does not work. For the DANRA domain as example that polygon only has points at the corners (blue), giving a weird cropping of lat-lon ERA data (red)

chull_points.png (view on web)<https://github.com/user-attachments/assets/87cecb56-b7a9-49cd-b4b9-e9a08690e7ee>

(It did speed up the cropping 10000 times though 😆). So we would need to also compute the distance to the actual edges of the polygon.

TomasLandelius:
Sorry, you had it right. Mixed up corners and edges...


-------- Originalmeddelande --------
Från: Landelius Tomas ***@***.***>
Datum: 2025-01-09 21:24 (GMT+01:00)
Till: mllam/mllam-data-prep ***@***.***>, mllam/mllam-data-prep ***@***.***>
Kopia: Mention ***@***.***>
Ämne: RE: [mllam/mllam-data-prep] Add support for cropping domain with padded lat/lon convex hull (PR #45)

You need to calculate the distance to the polygon, not just it's edges. It should be possible by considering all lines (great arcs) between the points making up the polygon.


-------- Originalmeddelande --------
Från: Joel Oskarsson ***@***.***>
Datum: 2025-01-09 18:08 (GMT+01:00)
Till: mllam/mllam-data-prep ***@***.***>
Kopia: Landelius Tomas ***@***.***>, Mention ***@***.***>
Ämne: Re: [mllam/mllam-data-prep] Add support for cropping domain with padded lat/lon convex hull (PR #45)


@joeloskarsson commented on this pull request.

________________________________

In mllam_data_prep/ops/cropping.py<https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1909178516>:

> +    def calc_dist(da_pt_xyz):
+        dotproduct = np.dot(da_xyz_ref, da_pt_xyz.T)
+        val = np.min(np.arccos(dotproduct))
+        return val
+
+    da_mindist_to_ref = xr.apply_ufunc(
+        calc_dist,
+        da_xyz,
+        input_core_dims=[["xyz"]],
+        output_core_dims=[[]],
+        vectorize=True,
+    )


Extracting the points that make up the convex hull polygon in spherical_geometry and computing the distance to those points does not work. For the DANRA domain as example that polygon only has points at the corners (blue), giving a weird cropping of lat-lon ERA data (red)

chull_points.png (view on web)<https://github.com/user-attachments/assets/87cecb56-b7a9-49cd-b4b9-e9a08690e7ee>

(It did speed up the cropping 10000 times though 😆). So we would need to also compute the distance to the actual edges of the polygon.

mllam_data_prep/ops/cropping.py
joeloskarsson: I implemented the distance to arcs method and it seems to be working. Have not done extensive testing though, only on DANRA domain. 

It is in this commit: https://github.com/joeloskarsson/mllam-data-prep/commit/7be2215af923774a6d95c0ee57a45501b5ae338a Since I do most computations using pure numpy, I thought that you might want to tweak it to use more xarray. Otherwise you can just cherry-pick that commit onto here. Note that I also changed the margin_width_degrees for the ERA example, as the current 0.2 degrees will not capture any boundary points.

leifdenby: thanks @joeloskarsson I will take a look!

observingClouds: > Working with this I find that the complexity of this does indeed create issues. I am unable to actually create any boundary datasets based on 0.25 degree ERA5 due to this taking too long. That's one case where both N and M are large.

I can confirm this. While looking into #54 and testing it with this branch my local machine takes forever to do the cropping.

leifdenby: I've merged your branch with the faster algorithm @joeloskarsson. All the tests pass and this does feel significantly snappier :) Thanks @joeloskarsson and @TomasLandelius !

mllam_data_prep/create_dataset.py
leifdenby: Great, with your 👍 I'm going to assume we can resolve this conversation :)

example.era5_cropped.yaml
joeloskarsson: Working with this a bit more I do actually find it convenient to be able to specify a config rather than an existing zarr, as it allows for building both the interior and boundary zarrs at the same time. The re-computation of the interior zarr to use for boundary cropping is not really a big problem, as nothing is really computed except for the coordinates and some meta-data.

mllam_data_prep/ops/cropping.py
leifdenby: Rather than implement the means the which cropping is applied (interior only, margin only or margin and interior) as `method` that must be selected with a specific string value (above I had suggested `convex_hull_interior`, `convex_hull_margin` and `convex_hull_interior_and_margin`) I decided it would be clearer to just use the margin-width to decide if the margin is included and add a boolean called `include_interior_points` to choose if the interior should be included too. With this the config is as follows:

```python
@dataclass
class ConvexHullCropping:
    """
    Define the method applied for cropping the spatial domain before writing
    the transformed output dataset. This is typically used when you want to
    create a dataset to provide data in a boundary around a limited-area
    domain.

    The cropping is done by creating a convex hull around the spatial
    coordinates of an *interior* dataset (this will typically be the
    "limited-area" domain when doing Limited Area Modelling) and then including
    all points that are within a margin of the convex hull boundary. In addition
    to including the points inside the defined margin around the convex hull,
    you can also include the points inside the convex hull of the interior
    dataset by setting the `include_interior` attribute to `True`.

    Attributes
    ----------
    margin_width_degrees: float
        The width (in degrees) of the margin applied to the convex hull
        boundary of the interior dataset used to define the cropping domain.
    interior_dataset_config_path: str
        The path to the configuration file for the dataset defining the interior domain
    include_interior_points: bool
        Whether to include the points inside the convex hull of the interior dataset
    """

    margin_width_degrees: float
    interior_dataset_config_path: str
    include_interior_points: bool = False
```

leifdenby: Feel free to reopen this discussion if you would like to discuss this further :)

joeloskarsson: I agree that that is cleaner, looks good

TomasLandelius:
Are you sure about the workings of shapely in 3D? I read that "The return value is a strictly two-dimensional geometry. All Z coordinates of the original geometry will be ignored."

/Tomas
________________________________________

joeloskarsson:
Agreed, and thanks for opening #83 issue to track that optimization! If every issue from above was resolved this should be good to go. There is only https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1885472797 that is not resolved, but I think maybe that is already fixed?

TomasLandelius:
After some discussion with GPT-5 I may have came up with a possible solution:

Since the hull is convex, we can implement a vectorized “all dot products ≥ 0” test against the hull’s great‑circle half‑spaces. That avoids Python loops entirely:
- Each hull edge defines a great‑circle plane (normal vector).
- A point is inside iff it lies on the “inside” side of all those planes.
- That reduces to a set of dot‑product checks, which NumPy can do in bulk.
Each hull edge defines a great‑circle plane. The “inside” of the polygon is the intersection of all those half‑spaces. By orienting the normals consistently (using a reference interior point), you guarantee that “inside” corresponds to non‑negative dot products. The final np.all(dots >= 0, axis=1) is fully vectorized: no Python loops, just one matrix multiplication.

See the attached example script for a comparison.

Please verify...

/Tomas

________________________________________

mllam_data_prep/ops/cropping.py
joeloskarsson: Is this fixed, and if so where? I found https://github.com/leifdenby/mllam-data-prep/commit/ba467183ec88959aa529fc0574daf1c82b5d1a1c, which seems to get rid of the renaming, but won't that just make it so that the resulting ERA5 zarr is incompatible with neural-lam?

leifdenby:
> [#45 (comment)](https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1885472797) still looks unresolved to me, please have a look at that thread. Otherwise I agree, after changelog + a docs readme this should be good to go 😄

you're right, sorry! I loose track of things in these comments on github sometimes. So full TODO is:

- [ ] ensure coordinates are correctly included in resulting dataset (appears there may be issues with ERA5): https://github.com/mllam/mllam-data-prep/pull/45#pullrequestreview-2504290079
- [ ] document new fields in README: https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1885459518
- [ ] clarify that notebook demonstrates how domain cropping is done, but doesn't show how to use it: https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1885458540
- [ ] update changelog

mllam_data_prep/ops/cropping.py
leifdenby: > I suppose you could argue that that is a neural-lam issue? And that we should not require the `x`/`y` naming there?

Yes, I don't think `neural-lam` should make this assumption, as you say, that would make this a `neural-lam` issue. It isn't quite correct however that `neural-lam` *assumes* `x` and `y` to be the names of the Cartesian coordinates, rather they default to `x` and `y` in the `BaseRegularGridDatastore` class (https://github.com/mllam/neural-lam/blob/main/neural_lam/datastore/base.py#L466), but `MDPDatastore` the Cartesian coordinates are actually parsed from the `mllam-data-prep` config used: https://github.com/mllam/neural-lam/blob/main/neural_lam/datastore/mdp.py#L109. (I would like to not read the stacked coordinates from the config though, but rather store them correctly in the produced zarr dataset, since I now found out how to correctly do that, see https://github.com/mllam/mllam-data-prep/pull/85)

> Is this fixed, and if so where? I found [leifdenby@ba46718](https://github.com/leifdenby/mllam-data-prep/commit/ba467183ec88959aa529fc0574daf1c82b5d1a1c), which seems to get rid of the renaming, but won't that just make it so that the resulting ERA5 zarr is incompatible with neural-lam?

Yes, I believe this is fixed (on the `mllam-data-prep` side at least). As in the config example no longer makes any mention of `x` and `y` and I have added a test in https://github.com/mllam/mllam-data-prep/pull/45/commits/a9921e130b140ee54ff3b35d8a40bb3a4b17ee07 which compares producing a dataset from era5 with and without cropping it, ensuring that the variables and coordinate names are the same in both (this ensure that if a dataset produced with `mllam-data-prep` without cropping contains coordinates called `latitude` and `longitude`, then a dataset that has also been cropped will contain the same). Also, as I mentioned above `MDPDatastore` reads the `mllam-data-prep` config to determine what the Cartesian grid coordinates are actually called. I would say that `MDPDatastore.get_xy()` needs to be changed though. What we actually mean is return the "regular grid coordinates" I think, rather than they must be called `x` and `y`, so maybe `MDPDatastore.get_xy()` should also use `MDPDatastore.CARTESIAN_COORDS` as the stacking/unstacking routines in `MDPDatastore` does, rather than hardcoding that these coordinates must be named `x` and `y`. What do you think? I would say that is actually a bug in `neural-lam` right now, as in, it should be able to train on datasets on a regular lat/lon grid, but currently that is not possible. If you agree I'd be happy to make a PR for `neural-lam` to fix `MDPDatastore.get_xy()`

leifdenby:
Ok, apart from the discussion on https://github.com/mllam/mllam-data-prep/pull/45#discussion_r1885472797 this is ready to go now :) Just need to see if what I've done to resolve that discussion makes sense for you.

pyproject.toml
```diff
@@ -38,6 +38,13 @@ dask-distributed = [
     "dask[distributed]>=2024.7.1",
     "bokeh!=3.0.*,>=2.4.2",
 ]
+latlon-domain-crop = [
+    "spherical-geometry>=1.3.2",
+    "cartopy>=0.23.0",
+    "gcsfs>=2024.10.0",
+    "jinja2>=3.1.4",
+]
```
joeloskarsson: Are these really optional as is? It seems I can't really import and use any part of mdp without these right now. I don't see any problem with making them part of the normal requirements.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     "isodate>=0.6.1",
     "requests>=2.31.0",
     "aiohttp>=3.9.3",
-    "dataclass-wizard>=0.29.2",
+    "dataclass-wizard==0.30.0",
```
observingClouds: Is this necessary to pin this specific version?

pyproject.toml
joeloskarsson: Yea I think it is true that many users would not need these tools, but making the dependencies truly optional (making sure no imports fail) might be more effort than it is worth. Up to you.

PR #56, opened 2025-01-30
bugfix and minor refactor in derived variables

leifdenby:
## Describe your changes

This PR refactors variables extraction routines in derivation of variables and fixes a bug in mapping from dataset variable name to kwarg name (previously the `dict` `keys` were used to select the variables from the source dataset, it should have been the `values`).

This PR also removes the special treatment of the longitude and latitude coordinates. With that I mean that with the config structure as it is now for the derived variables you can simply define through the kwarg mapping whether your longitude variable is called `lon` or `longitude`. Using CF-convention attributes to find out what variables to use for latitude and longitude we will leave to #33 

No change in dependencies for this fix

## Issue Link

No issue, was discussed on slack

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [x] I have requested a reviewer and an assignee (assignee is responsible for merging)

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

- [ ] I have added a line to the CHANGELOG describing this change, in a section

## Checklist for assignee

- [ ] PR is up to date with the base branch
- [ ] the tests pass
- [ ] author has added an entry to the changelog (and designated the change as *added*, *changed* or *fixed*)

### DIFF
--- a/mllam_data_prep/ops/derive_variable/main.py
+++ b/mllam_data_prep/ops/derive_variable/main.py
@@ -45,31 +45,21 @@ def derive_variable(ds, derived_variable, chunking):
     function_namespace = derived_variable.function
     expected_field_attributes = derived_variable.attrs
 
-    # Arguments to the function to be selected/extracted from the input dataset
-    ds_kwargs = {
-        key: val.rpartition(".")[2]
-        for key, val in derived_variable.kwargs.items()
-        if "ds_input" in val
-    }
-    # Other arguments that should not be selected/extracted form the input dataset
-    other_kwargs = {
-        key: val
-        for key, val in derived_variable.kwargs.items()
-        if "ds_input" not in val
-    }
+    # split the function kwargs defined in the config into two groups:
+    # 1. variables that should be extracted from the input dataset (and renamed)
+    # 2. other kwargs that should be passed in as is
+    required_input_dataset_vars = {}
+    other_kwargs = {}
+    for key, val in derived_variable.kwargs.items():
+        if val.startswith("ds_input."):
+            var_name = val.rpartition(".")[2]
+            required_input_dataset_vars[key] = var_name
+        else:
+            other_kwargs[key] = val
 
-    # Separate the lat,lon from the required variables as these will be derived separately
-    logger.warning(
-        "Assuming that the lat/lon coordinates are given as variables called"
-        " 'lat' and 'lon'."
-    )
-    latlon_coords_to_include = {}
-    for key, val in list(ds_kwargs.items()):
-        if val in ["lat", "lon"]:
-            latlon_coords_to_include[key] = ds_kwargs.pop(key)
-
-    # Get subset of input dataset for calculating derived variables
-    ds_subset = ds[ds_kwargs.keys()]
+    # select from the input dataset the subset of variables which have been
+    # selected to be used as input arguments for the derived variable
+    ds_subset = ds[list(required_input_dataset_vars.values())]
 
     # Chunking is needed for coordinates used to derive a variable since they are
     # not lazily loaded, as otherwise one might run into memory issues if using a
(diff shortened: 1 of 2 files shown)


mllam_data_prep/ops/derive_variable/main.py
```diff
@@ -45,31 +45,21 @@ def derive_variable(ds, derived_variable, chunking):
...
+            var_name = val.rpartition(".")[2]
+            required_input_dataset_vars[key] = var_name
+        else:
+            other_kwargs[key] = val
 
-    # Get subset of input dataset for calculating derived variables
-    ds_subset = ds[ds_kwargs.keys()]
+    # select from the input dataset the subset of variables which have been
+    # selected to be used as input arguments for the derived variable
+    ds_subset = ds[list(required_input_dataset_vars.values())]
```
ealerskans, line 62: Thanks for the fix! I must have forgotten to change this when I switched from using the `keys` to the `values` as the named argument to the function

mllam_data_prep/ops/derive_variable/main.py
```diff
@@ -45,31 +45,21 @@ def derive_variable(ds, derived_variable, chunking):
...
-
-    # Separate the lat,lon from the required variables as these will be derived separately
-    logger.warning(
-        "Assuming that the lat/lon coordinates are given as variables called"
-        " 'lat' and 'lon'."
-    )
-    latlon_coords_to_include = {}
-    for key, val in list(ds_kwargs.items()):
-        if val in ["lat", "lon"]:
-            latlon_coords_to_include[key] = ds_kwargs.pop(key)
```
ealerskans, line 69: I thought the idea was to remove the latitude and longitude variables from the `dict` with variables to be extracted/selected from the input dataset (`required_input_dataset_vars`). This since we wanted to have a separate way of getting them, depending on if they are in the dataset or if they should be generated from the projection information.

It is as you say now possible to extract them directly from the dataset by specifying the correct variable name (`lat`, `latitude`, etc.). So I don't have any strong feeling about if this is removed or not.

ealerskans (commented):
Thanks for the bug fixes and refactoring!

I only have a minor comment/question about the latitudes and longitudes. I don't have any strong feeling about if we remove that part or not. Otherwise I am happy to approve this :)

mllam_data_prep/ops/derive_variable/main.py
leifdenby: Yes, that was my plan originally. But am realising that if we want to treat latitude and longitude (or just grid coordinates in general) in a special way, by assuming that they are defined with specific names or attributes in the dataset, then it is probably best to simply use the CF-conventions which cover this too (rather than coming up with our own naming convention). Until we implement that (which is quite a lot more work) I think sticking with just treating these as variables/coordinates that have to be explicitly named is safest. At least it is explicit :D

mllam_data_prep/ops/derive_variable/main.py
leifdenby: no problem! I'm glad I hadn't misunderstod and you agree this is how it should work

mllam_data_prep/ops/derive_variable/main.py
ealerskans: Sounds good :)

ealerskans (approved):

Issue #58, opened 2025-01-31
Issue when adding coordinates to the output dataset


ealerskans:
I was having issues generating a dataset from the full DANRA pressure levels data. Everything works nicely with the small test dataset in the example config (https://mllam-test-data.s3.eu-north-1.amazonaws.com/height_levels.zarr) but I was getting a lot of nans in my output when I was using the full dataset. The reason for this is apparently this part here

https://github.com/mllam/mllam-data-prep/blob/3a48c997144e66ef2478413668e669e124400cdd/mllam_data_prep/create_dataset.py#L157-L161

more specifically the part where the dataset to be saved to zarr (`ds`) is assigned coordinates (lines 160-161). For the full DANRA pressure levels dataset this means that we will assign the following coordinates to `ds`

```
Coordinates:
  * time      (time) datetime64[ns] 774kB 1990-09-01 ... 2023-10-13T21:00:00
  * pressure  (pressure) int64 112B 1000 950 925 900 850 ... 400 300 250 200 100
  * y         (y) float64 5kB -6.095e+05 -6.07e+05 ... 8.58e+05 8.605e+05
  * x         (x) float64 6kB -1.999e+06 -1.997e+06 ... -3.175e+04 -2.925e+04
```

and if we then e.g. want to select the 1000 hPa `z` variable this would be a data-array, `da`, with only have 1 "pressure  dimension"
```
<xarray.DataArray 'z' (time: 96768, pressure: 1, y: 589, x: 789)> Size: 360GB
dask.array<getitem, shape=(96768, 1, 589, 789), dtype=float64, chunksize=(256, 1, 295, 263), chunktype=numpy.ndarray>
Coordinates:
    lat       (y, x) float64 4MB dask.array<chunksize=(295, 263), meta=np.ndarray>
    lon       (y, x) float64 4MB dask.array<chunksize=(295, 263), meta=np.ndarray>
  * pressure  (pressure) int64 8B 1000
  * time      (time) datetime64[ns] 774kB 1990-09-01 ... 2023-10-13T21:00:00
  * x         (x) float64 6kB -1.999e+06 -1.997e+06 ... -3.175e+04 -2.925e+04
  * y         (y) float64 5kB -6.095e+05 -6.07e+05 ... 8.58e+05 8.605e+05
```

[...]

PR #60, opened 2025-02-04
Fix unwanted dimensions in dataset

ealerskans:
## Describe your changes

This PR fixes a bug in what dimensions are added to the output dataset. Currently,  the dimensions of the input dataset are all added. However, if one only wants to extract e.g. `u` and `v` from the 100 m altitude in the DANRA height levels dataset, the current implementation will add `altitude` as a dimension meaning that all selected variables will be broadcasted to have `altitude` as a dimension, even though we're only selecting one variables from one altitude. This is fixed by removing the coordinates assignment to the output dataset when initialising it.

Furthermore, there is a similar bug for the derived variables, in which the input dataset dimensions are used. The change by @joeloskarsson (https://github.com/joeloskarsson/mllam-data-prep/commit/8040378a72b5d9740d176f023eede876a7b67739) fixes this by using the expected input variable dimensions instead as `target_dims`.

No change in dependencies for this fix. 

## Issue Link

Tries to solve #58.

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [x] I have requested a reviewer and an assignee (assignee is responsible for merging)

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented (including return types and parameters)
- [x] the code is easy to maintain

- [x] I have added a line to the CHANGELOG describing this change, in a section

## Checklist for assignee

- [x] PR is up to date with the base branch
- [x] the tests pass
- [x] author has added an entry to the changelog (and designated the change as *added*, *changed* or *fixed*)

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1,12 +1,11 @@
 """Tests for the output dataset created by `mllam-data-prep`."""
-import re
-import shutil
-import tempfile
-from pathlib import Path
-
 import pytest
+import yaml
 
 import mllam_data_prep as mdp
+
+with open("example.danra.yaml", "r") as file:
+    BASE_CONFIG = file.read()
 
 HEIGHT_LEVEL_TEST_SECTION = """\
 inputs:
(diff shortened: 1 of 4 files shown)


observingClouds (commented):
Thanks for the PR @ealerskans! Could you please add some tests, e.g. that the output variables are the one selected and that there are no NaN values?

tests/test_dataset.py
```diff
@@ -0,0 +1,301 @@
...
+        dim: time
+      state_feature:
+        method: stack_variables_by_var_name
+        dims: [pressure]
+        name_format: "{var_name}{pressure}m"
+      grid_index:
+        method: stack
+        dims: [x, y]
+    target_output_variable: state
+"""
```
ealerskans, line 273: This part here is related to issue #61. We agreed that we should be able to support this kind of selection, where we select variable form different levels. But this is currently not supported and would require re-writing of the code. So not sure if this PR should be merged before that has been addressed. I am in process of implementing this and will hopefully create a PR for it later today.

ealerskans: Actually, I am now in doubt if this would even work after that change since that will introduce a check in the reading of the config file and then throw an error if we're trying to make this kind of selection.

observingClouds: This is a tricky question. I would still implement all tests and then add a decorator to the test that will allow it to be skipped in case a NotImplementedError is raised, like [here](https://stackoverflow.com/a/63522579) but specific to `NotImplementedError`

ealerskans: As agreed, I will implement the decorator addition in #65 instead :)

observingClouds:
blocked by #66 / #67

leifdenby:
no longer blocked now that @observingClouds has merged #66 and #67 🥳

observingClouds (approved):
Sweet! This looks good to me now 😉

